PR #2, opened 2025-01-09
Fix `ValueError: high is out of bounds for int32` in `randint` Calls

Horikitasaku:
This PR fixes a `ValueError` that occurs when using `rnd.randint` with values exceeding the bounds of `int32`. Specifically, the error occurs because the `high` parameter in `rnd.randint` defaults to `int32`, causing issues when `2**32 - 1` is used. 

The fix ensures compatibility by explicitly setting the `dtype` to `np.int64`.

---

### Changes Made

- Updated calls to `rnd.randint` to include `dtype=np.int64`, as shown below:

**Before:**
```python
rnd.randint(0, 2**32 - 1)
```

**After:**
```python
rnd.randint(0, 2**32 - 1, dtype=np.int64)
```

---

### Reproduced Error

The error message observed is as follows:
```
ValueError: high is out of bounds for int32
```

**Stack Trace:**
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
... (truncated for brevity) ...
File numpy\\\\random\\\\mtrand.pyx:780, in numpy.random.mtrand.RandomState.randint()
File numpy\\\\random\\\\_bounded_integers.pyx:2881, in numpy.random._bounded_integers._rand_int32()
```

---

### Impacted Files

The change affects the following files:

- `src/tabpfn_extensions/hpo/tuned_tabpfn.py`
- `src/tabpfn_extensions/post_hoc_ensembles/abstract_validation_utils.py`
- `src/tabpfn_extensions/post_hoc_ensembles/sklearn_interface.py`
---

### Testing

- Verified that the updated code runs without errors.
- Ensured reproducibility by maintaining random state initialization across different configurations.

---

### Checklist

- [x] Updated code to fix the error.
- [x] Verified that the changes do not break existing functionality.
- [x] Added this PR description to document the issue and resolution clearly.

---

### Additional Notes

If you encounter any issues or have suggestions for further improvement, please feel free to comment on this PR.

### DIFF
--- a/src/tabpfn_extensions/hpo/tuned_tabpfn.py
+++ b/src/tabpfn_extensions/hpo/tuned_tabpfn.py
@@ -87,8 +87,8 @@ class TunedTabPFNBase(BaseEstimator):
         rng = check_random_state(self.random_state)
 
         # Set random seeds for reproducibility
-        torch.manual_seed(rng.randint(0, 2**32 - 1))
-        np.random.seed(rng.randint(0, 2**32 - 1))
+        torch.manual_seed(rng.randint(0, 2**32 - 1,dtype=np.int64))
+        np.random.seed(rng.randint(0, 2**32 - 1,dtype=np.int64))
 
         # Fit transformers
         X = self._cat_encoder.fit_transform(X)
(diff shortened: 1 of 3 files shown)


LennartPurucker:
Heyho, thank you for the fix, but I suggest a slightly different, potentially safer route to fix this. 

I proposed the fix in the following PR: #3 
Happy to hear your thoughts about it.

PR #3, opened 2025-01-09
fix: out of bounds int32 in randint calls

LennartPurucker:
Fixes max int bound for randint calls. See #2 for more details.


eddiebergman (approved):

Horikitasaku:
This is better :)

PR #9, opened 2025-01-14
Fix phe and hpo with TabPFN API

LeoGrin:
The PHE and HPO break in a bunch of way with the client. This PR (jointly with a [server PR](https://github.com/automl/tabpfn-server/pull/135) and a [client PR](https://github.com/PriorLabs/tabpfn-client/pull/75) fix these issues. This `tabpfn-extensions` PR:
- import PreprocessorConfig from the right place
- fix the wrapper to make it work (you need to pass all the params in the constructor) and to support `categorical_features_indices` (the client doesn't use it for now so we just output a warning)

### DIFF
--- a/src/tabpfn_extensions/__init__.py
+++ b/src/tabpfn_extensions/__init__.py
@@ -6,6 +6,6 @@ except PackageNotFoundError:
     __version__ = "0.1.0.dev0"
 
 from .utils import is_tabpfn
-from .utils import TabPFNRegressor, TabPFNClassifier, PreprocessorConfig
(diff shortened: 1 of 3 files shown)


noahho:
I like the change as this moves some more of the shared funcs into the common package! Would you move things from the main TabPFN package into the shared utils too?

noahho (approved):

noahho:
This still works with the non-client, i.e. with the local package? If yes LGTM

LeoGrin:
Waiting for the new tabpfn package release to merge.

PR #15, opened 2025-01-16
Remove | to make python3.9 compatible

### LINKED ISSUE #10: Issues with python3.9
python3.9 is supposed to be supported in the pytoml but some part seems to be failing, for instance:
```
RandomForestTabPFNClassifier | RandomForestTabPFNClassifier,
TypeError: unsupported operand type(s) for |: 'ABCMeta' and 'ABCMeta'
```
from https://github.com/PriorLabs/TabPFN/issues/127

LeoGrin:
Fix #10 
This one wasn't caught by future annotations because it's not in type hints.
I also think there was an error and it was supposed to be RFTab `RandomForestTabPFNClassifier | RandomForestTabPFNRegressor` instead of `RandomForestTabPFNClassifier | RandomForestTabPFNClassifier`. @noahho I think you wrote this, could you check this is correct?

### DIFF
--- a/src/tabpfn_extensions/post_hoc_ensembles/pfn_phe.py
+++ b/src/tabpfn_extensions/post_hoc_ensembles/pfn_phe.py
@@ -374,7 +374,7 @@ class AutoPostHocEnsemblePredictor(BaseEstimator):
         for _, bm in bm_list:
             if isinstance(
                 bm,
-                RandomForestTabPFNClassifier | RandomForestTabPFNClassifier,
+                (RandomForestTabPFNClassifier, RandomForestTabPFNRegressor),
             ):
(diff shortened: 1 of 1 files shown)


src/tabpfn_extensions/post_hoc_ensembles/pfn_phe.py
```diff
@@ -374,7 +374,7 @@ def _collect_base_models(
         for _, bm in bm_list:
             if isinstance(
                 bm,
-                RandomForestTabPFNClassifier | RandomForestTabPFNClassifier,
+                (RandomForestTabPFNClassifier, RandomForestTabPFNRegressor),
```
noahho, line 377: Ohh, thanks for double-catchign this one

noahho (approved):
LGTM

PR #23, opened 2025-01-29
Fix float32 error in RandomForestTabPFN

LeoGrin:
Fix https://github.com/PriorLabs/TabPFN/issues/162, which would appear when X contains NaNs and scikit-learn<1.6 + small fixes

https://github.com/PriorLabs/TabPFN/issues/162 is fixed by removing the `_validate_data` override from RandomForest. I also removed other overrides which do not seem useful.

I tried the scikit-learn `check_estimator` tests, and a lof them are failing for RandomForestTabPFN, but I that'll be a bigger PR!


LennartPurucker (approved):
LGTM

PR #24, opened 2025-02-01
Rename LICENCE file to be compatible with dependency managment tools

fif911:
The current issue is that this package cannot be installed via dependency management tools like poetry because the LICENCE file is named incorrectly.

For example:

```poetry add git+https://github.com/priorlabs/tabpfn-community.git```

Fails with  "No such file or directory: .../src/tabpfn-community/LICENSE"

This small PR resolves the issue.

Also I am not sure why there is a hard dependency on Seaborn version. Maybe we can remove it or make it less specific for better compatibility.


noahho (approved):

noahho:
Thanks a lot, will also change in main repo

PR #28, opened 2025-02-16
Update README.md

mert-kurttutan:
Add more convenient pip install from git

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,10 +14,10 @@
 
 ```bash
 # Clone and install the repository
-!git clone https://github.com/PriorLabs/tabpfn-community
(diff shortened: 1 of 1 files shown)


mert-kurttutan:
Small change but, I found this more convenient when running code with community extensions, no need to deal with artifact of source code in local directories

LeoGrin:
Thanks @mert-kurttutan, it also seems more convenient to me! I'm actually thinking we could invert the two and make your solution the default. Also, the current solution could be improved (remove the "!" and add "" around the `tabpfn-community[post_hoc_ensembles,interpretability,hpo]`)

LeoGrin:
Thanks @mert-kurttutan, the cloning should only be there for the second option though

PR #30, opened 2025-02-18
Always use safepower instead of power to fix bracket errors

LeoGrin:
Fix https://github.com/PriorLabs/TabPFN/issues/183
I'm wondering why we didn't always use `safepower` before. For instance one of the search spaces contained both `safepower` and `power`, which is a bit confusing for me. So not 100% sure this fix is the right one.

Note that all of this will be fixed once our sklearn versions use scipy's yeo-johnson optimizer https://github.com/scikit-learn/scikit-learn/issues/26308


LennartPurucker (approved):
LGTM

PR #39, opened 2025-03-08
Add K-fold Embedding for TabPFN v2

6sy666:
The embedding extraction method described in [A Closer Look at TabPFN v2: Strength, Limitation, and Extension](https://arxiv.org/abs/2502.17361). Specifically:

- When extracting embeddings for the training set, I split the training set into multiple folds. Each fold is treated as a **query set**, while the remaining folds serve as the **support set**. This allows the extraction of embeddings for each fold separately.
- The embedding extraction for the test set remains the same as in the original implementation.
- With this method, the embeddings of the training set and the test set become more comparable.

### DIFF
--- a/examples/embedding/README.md
+++ b/examples/embedding/README.md
@@ -1,0 +1,30 @@
+# TabPFN Embeddings
+
+The `TabPFNEmbedding` class is a utility for extracting embeddings from `TabPFNClassifier` or `TabPFNRegressor` models. It supports both standard training (vanilla embeddings) and K-fold cross-validation for embedding extraction. The embeddings can be used to enhance the performance of downstream machine learning models.
+
+## Key Features
+
+- **Vanilla Embeddings**: Extract embeddings by training on the entire dataset when `n_fold=0`.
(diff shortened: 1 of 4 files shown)


noahho:
Great work!
You could also add a README file in embeddings/README.md referencing your paper:

```bibtext
@misc{ye2025closerlooktabpfnv2,
      title={A Closer Look at TabPFN v2: Strength, Limitation, and Extension}, 
      author={Han-Jia Ye and Si-Yang Liu and Wei-Lun Chao},
      year={2025},
      eprint={2502.17361},
      archivePrefix={arXiv},
      primaryClass={cs.LG},
      url={https://arxiv.org/abs/2502.17361}, 
}
```

Otheriwse looks good to me!

6sy666:
> Great work! You could also add a README file in embeddings/README.md referencing your paper:
> 
> ```
> @misc{ye2025closerlooktabpfnv2,
>       title={A Closer Look at TabPFN v2: Strength, Limitation, and Extension}, 
>       author={Han-Jia Ye and Si-Yang Liu and Wei-Lun Chao},
>       year={2025},
>       eprint={2502.17361},
>       archivePrefix={arXiv},
>       primaryClass={cs.LG},
>       url={https://arxiv.org/abs/2502.17361}, 
> }
> ```
> 
> Otheriwse looks good to me!

Thank you for your feedback! I’ve added the README.md file in the embeddings/ directory, as suggested, and included the reference to our paper.

PR #67, opened 2025-04-14
Pass min_samples_leaf to decision tree constructor

dev-rinchin:


noahho:
@dev-rinchin Thanks a lot for the contribution and PR! Could you briefly sketch out what the problem is this PR solved and sign the CLA? Then we're ready to go :-)

noahho:
Great, that makes sense. Thanks a lot your the contribution!

PR #75, opened 2025-05-05
Add extension simulation

davidotte:
Add function decorator @simulate_first, which simulates any function first by overwriting the fit and predict functions of TabPFN and simulating the actual execution time (and credit usage for the client). If the client is used and the expected credit usage surpasses the available credits, the function is not executed. I left some comments in the code which hopefully explain how everything works. So far, I only added the decorator to the PHE, please let me know to which other extensions I should add it.


noahho:
@davidotte please fix failing tests

noahho (changes requested):
@davidotte please fix failing tests. Also there seem a lot of changes to files that are unrelated to your PR, please remove those changes or explain why?

davidotte:
@noahho As discussed, tests fail because of min python version specified in tabpfn-common-utils, will fix this soon. To keep the code clean, I decided to create a utils folder and move utils.py as well as my simulation file into this folder. Therefore I needed to update the usage of utils to the new path, this is why I slightly changed some of the other files. If you think another solution would be nicer, let me know.

src/tabpfn_extensions/utils/utils.py
```diff
@@ -254,8 +254,8 @@ def get_params(self, deep: bool = True) -> dict[str, Any]:
             return params
 
 except ImportError:
-    TabPFNClassifierWrapper = None
-    TabPFNRegressorWrapper = None
+    ClientTabPFNClassifier = None
```
noahho, line 257: Why these changes?

src/tabpfn_extensions/utils/utils.py
```diff
@@ -287,8 +287,8 @@ def get_tabpfn_models() -> tuple[type, type]:
         logging.info("Using TabPFN package")
 
         return LocalTabPFNClassifier, LocalTabPFNRegressor
-    elif TabPFNClassifierWrapper is not None:
-        return TabPFNClassifierWrapper, TabPFNRegressorWrapper
+    elif ClientTabPFNClassifier is not None:
```
noahho, line 290: Why these changes?

noahho:
I'm not a fan of the utils.utils import path. Could we add things to the __init__.py file or import there?

src/tabpfn_extensions/utils/utils.py
davidotte: Just renamed it for clarity and to be consistent with the naming of the local models, e.g. LocalTabPFNClassifier. I had struggle understanding this part first, which is why I changed it on the fly. But you're right, this doesn't necessarily belong to this PR, I can also remove it if you prefer that.

src/tabpfn_extensions/utils/utils.py
davidotte: same as above

PR #76, opened 2025-05-09
Fix Many Class Codebook generation and tests

noahho:


MaxSchambach:
@noahho thank you! Indeed it seems to have resolve the problem in terms of errors being raised 🙌  
I can't yet say with regard to performance, will have to run and compare more experiments :)

noahho:
@MaxSchambach great! As I said, the classes range seems far outside of what TabPFN was made for so I don't expect anything strong here. Curious nevertheless :-)

PR #93, opened 2025-06-11
Support for DAG in unsupervised synthesis / imputation

marco-virgolin-ist:
Hi,

This PR is includes the support for causal DAGs in synthetic data generation/imputation.
DAGs are expressed as a dictionary `{ int: list[int] }` where the key is the depenent column_idx and the value is a list of column indices the key depends on.
E.g. for column 0 independent, column 1 depending on 0, column 2 depending on 1 and 0, you'd have:
```
{ 0: [], 1: [0], 2: [1,0] }
```

The PR modifies (IMHO minimally) `src/tabpfn_extensions/unsupervised/unsupervised.py` and adds a respective `examples/unsupervised/generate_data_following_dag.py`. Some small changes apply to other files.

Regarding the `src` file:
- enables to pass the dag dictionary to appropriate function calls.
- read the dag dictionary in `impute_` and order the variables using python's `graphlib TopologicalSorter` to synthesize them in the right order (e.g. independent first)
- set the `conditional_idx` of the `column_idx` being generated/imputed to its dependencies (e.g., if `column_idx==2` then `conditional_idx==[0,1]`)
- proceed as before

What this PR does **not** do (based on the contribution guidelines):
- does not create a new extension and applicable aspects
- does not include tests (!!!) Unfortunately I see no tests for `unsupervised` to add to (am I missing something?). I hope the limited scope of the proposed changes + the provided example suffice. If you feel otherwise, it would be great if a test suite for `unsupervised` is created, to which we could then add tests specific to the addition of dag support.

### DIFF
--- a/examples/unsupervised/generate_data_following_dag.py
+++ b/examples/unsupervised/generate_data_following_dag.py
@@ -1,0 +1,79 @@
+#  Copyright (c) Prior Labs GmbH 2025.
+#  Licensed under the Apache License, Version 2.0
+
+import torch
+from sklearn.datasets import load_wine
+from sklearn.model_selection import train_test_split
+
+from tabpfn_extensions import TabPFNClassifier, unsupervised
+
+# Load the breast cancer dataset
+df = load_wine(return_X_y=False)
+X, y = df["data"], df["target"]
+attribute_names = df["feature_names"]
+
+wine_dag = {
+    'alcohol': [],
+    'malic_acid': [],
+    'ash': ['magnesium'],
+    'alcalinity_of_ash': ['ash', 'magnesium'],
+    'magnesium': [],
+    'total_phenols': ['flavanoids', 'nonflavanoid_phenols', 'proanthocyanins'],
+    'flavanoids': [],
+    'nonflavanoid_phenols': [],
+    'proanthocyanins': [],
+    'color_intensity': ['flavanoids', 'proanthocyanins', 'total_phenols'],
+    'hue': ['color_intensity'],
+    'od280/od315_of_diluted_wines': ['flavanoids', 'total_phenols'],
+    'proline': ['alcohol', 'total_phenols']
+}
+
+# convert feature names to indices in keys and values
+dag = {i: [list(wine_dag.keys()).index(dep) for dep in deps] for i, deps in enumerate(wine_dag.values())}
+print(dag)
+
+# Split the data
+X_train, X_test, y_train, y_test = train_test_split(
+    X,
+    y,
+    test_size=0.5,
+    random_state=42,
+)
+
+# Initialize TabPFN models
+# Use parameters that work with both TabPFN and TabPFN-client
+clf = TabPFNClassifier(n_estimators=3)
+
+# Import TabPFNRegressor for numerical features
+from tabpfn_extensions import TabPFNRegressor
+
+reg = TabPFNRegressor(n_estimators=3)
+
+# Initialize unsupervised model
+model_unsupervised = unsupervised.TabPFNUnsupervisedModel(
+    tabpfn_clf=clf,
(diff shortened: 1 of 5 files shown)


noahho:
Hi @marco-virgolin-ist — owe you a real apology. We didn't have a clear internal answer on whether DAG-conditioned synthesis was something we wanted to commit to maintaining inside `tabpfn-extensions`, and rather than tell you that, I left this sitting. That's on me, not on the quality of your contribution.

I just pushed a rebase + small cleanup pass onto your branch (with you and @DavideTugnoli credited as `Co-authored-by:` on the commit, attribution preserved). What changed vs. your original:

- Rebased onto current `main` (your branch base was 11 months stale, so it was unmergeable as-is)
- `CycleError` → `ValueError("DAG contains a cycle: …")` so users debugging their own DAGs get a friendly message instead of a stdlib traceback
- Built a local copy of the DAG dict instead of mutating the caller's input
- Extracted the conditioning-mode selection into a small helper so the three modes (DAG / sequential / all-features) live in one place
- Added a test file (`tests/test_unsupervised_dag.py`) — topological order, partial DAG, cycle detection, no-caller-mutation, mode mutex, and an end-to-end synthesis run in `FAST_TEST_MODE`
- Tightened the example (the `breast cancer` comment was a copy-paste leftover from `load_breast_cancer`; switched it to `load_wine` consistently and labeled the DAG as illustrative)

**Mechanically the PR is now in a state we could merge today.** Before we do, could you tell us a bit about the use case driving this? We're trying to figure out whether DAG-conditioned synthesis is a feature with users beyond your own work — that determines whether we commit to maintaining it as a first-class part of `tabpfn-extensions`, or whether the cleaner home is a small standalone package linked from [`awesome-tabpfn`](https://github.com/PriorLabs/awesome-tabpfn).

Either answer is fine, but the use-case detail is what'd unblock the decision.

Sorry again for the wait, and thanks for the patience.

DavideTugnoli:
Hi @noahho, and thanks for the cleanup. It all looks good to me.

The main use case is generating synthetic data when real-world data is scarce but (partial) causal knowledge is available, which is common in medical and pharma settings. As a concrete example, InSilicoTrials, a company we collaborate with, applies this in a large pharma engagement, where synthetic patient data feeds causal inference and the data carries confounders, mediators, and colliders rather than a single exposure-outcome pair. Conditioning on the known causal structure enforces a generation order that vanilla sampling does not, which helps avoid biased estimates and spurious correlations. They obtain the graph in several ways (stakeholder knowledge, causal discovery, and LLM-from-metadata refined by a user), the client asked for workshops on it, and they are extending it to further indications, so the need is real and recurring.

The reason it helps is that TabPFN's autoregressive generation is sensitive to the column ordering, and the permutation averaging reduces but does not remove this. In our paper, accepted at UAI 2026, we explain why this happens and show that causal conditioning improves both the distributional quality of the synthetic data and treatment-effect preservation: https://arxiv.org/abs/2603.10254

Note: the current code expects a complete DAG. For partial knowledge, in the paper we use a hybrid rule that conditions on the trusted edges and falls back to sequential conditioning for the rest. There are several reasonable ways to design this, and it is not part of this PR.

Thank you,
Davide, Giovanni (@Giovannicina), and Marco (now @marcovirgolin; the @marco-virgolin-ist account is no longer active)

noahho:
Continued in #306 — opened on a Prior Labs branch since this PR's fork has `maintainerCanModify` off and @marco-virgolin-ist's account is inactive. Co-author trailers on the commit preserve attribution. Same content, ruff fixed, slightly tighter; ready for review/merge there.

marcovirgolin:
thank you @noahho  and @DavideTugnoli for continuing this effort.

PR #94, opened 2025-06-13
Make data shuffling optional in _optimize function for HPO

klemens-floege:
Make data shuffling optional in _optimize function for HPO through introducing a new parameter in the base class

### DIFF
--- a/src/tabpfn_extensions/hpo/tuned_tabpfn.py
+++ b/src/tabpfn_extensions/hpo/tuned_tabpfn.py
@@ -104,6 +104,7 @@ class TunedTabPFNBase(BaseEstimator):
         self,
         n_trials: int = 50,
         n_validation_size: float = 0.2,
+        shuffle_data: bool = False,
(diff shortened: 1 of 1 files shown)


noahho (approved):

src/tabpfn_extensions/hpo/tuned_tabpfn.py
```diff
@@ -104,6 +104,7 @@ def __init__(
         self,
         n_trials: int = 50,
         n_validation_size: float = 0.2,
+        shuffle_data: bool = False,
```
noahho: DEAFULT MUST BE TRUE

noahho:
Before merging: shuffling default must be true!

src/tabpfn_extensions/hpo/tuned_tabpfn.py
klemens-floege: Agreeee!

PR #95, opened 2025-06-13
[WIP] post_hoc_ensembling with custom cross-validation splits

klemens-floege:
Implement code custom cross-validation splits into AutoTabPFN while retaining original functionality if set to None. Also an example script to demonstrate usage, but I think it would be better to merge this into `phe_example.py` with a simpler dataset.


klemens-floege:
outdated

PR #97, opened 2025-06-19
Better setup

noahho:


scripts/get_max_dependencies.py
```diff
@@ -15,7 +15,7 @@ def main() -> None:
     deps_match = re.search(r"dependencies\s*=\s*\[(.*?)\]", content, re.DOTALL)
     if deps_match:
         deps = [
-            d.strip(' "\'')
```
LeoGrin, line 18: is it working or is it ruff being annoying?

src/tabpfn_extensions/hpo/tuned_tabpfn.py
```diff
@@ -285,15 +285,23 @@ def objective(params):
                 else:  # Regression
                     y_pred = model.predict(X_val)
                     if self.metric == MetricType.RMSE:
-                        score = -mean_squared_error(
-                            y_val, y_pred, squared=False
-                        )  # Negative RMSE
+                        score = -score_regression(
```
LeoGrin, line 280: I would say we can leave it as before, or pass the metric directly to score_regression.

src/tabpfn_extensions/hpo/tuned_tabpfn.py
```diff
@@ -373,28 +381,10 @@ def objective(params):
         self.trials_ = trials_obj  # Store the modified trials history
 
         if self.best_model_ is None:
-            warnings.warn(
-                "All optimization trials failed or no valid model was found. Creating default model.",
-                stacklevel=2,
+            raise ValueError(
```
LeoGrin, line 380: could it be an issue in AutoTabPFNClassifier if this happens from time to time but is fine as part of the PHE? (No idea if that's actually a thing)

LeoGrin (approved):
A few comments, otherwise LGTM!

LeoGrin:
bugbot run

scripts/get_max_dependencies.py
noahho: took the tabpfn workflow

src/tabpfn_extensions/hpo/tuned_tabpfn.py
noahho: the way before crashed on some sklearn versions whoch is why we change but you're right to keep consistent

src/tabpfn_extensions/hpo/tuned_tabpfn.py
noahho: this is independent of the AutoTabPFN Code

PR #108, opened 2025-07-08
RegClass implementation

klemens-floege:
Implemention of Regressor-as-classifier. With synthetic data example for binary and mutliclass classification.


klemens-floege:
outdated

PR #110, opened 2025-07-09
AutoGluon Tabular

klemens-floege:
AutoGluon Tabular implementation based on TabArena 

TODOs:
- Test cases
- Fix imports

### DIFF
--- a/src/tabpfn_extensions/autogluon/sklearn_interface.py
+++ b/src/tabpfn_extensions/autogluon/sklearn_interface.py
@@ -5,7 +5,8 @@ from typing import Literal
 import numpy as np
 import pandas as pd
 
-
+# TODO: Same interface as AutoTabPFN
+# Identical to before
 class _BaseAutoGluonTabPFN:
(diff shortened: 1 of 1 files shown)


LennartPurucker:
FYI: I refactored something small(ish) for the TabArena code today: https://github.com/autogluon/tabrepo/pull/183

Ideally, we will get this also into the PR (with DF-PFN) for AG: https://github.com/autogluon/autogluon/pull/5191 
If that gets merged, you could just bootstrap on that version and would not need to add/maintain the other abstract model files.

Also, here is an example for using the PHE versions of models from TabArena via AutoGluon: https://github.com/TabArena/tabarena_benchmarking_examples/blob/main/tabarena_minimal_example/running_tabarena_models/run_tuned_ensemble_model.py

Also, with AG 1.4 (to be released soon) we are fixing some GPU-related bugs that should make running your sklearn interface a better experience.

noahho (changes requested):
RE our conversation

PR #140, opened 2025-08-14
Remove Buggy Check for samples + features

klemens-floege:
Bug Fix for faulty n_feaures check in AutoTabPFN. Set presets default to "best_quality"- 

Tested for 5k dataset, and it worked well with max_time = 6 minutes

### DIFF
--- a/src/tabpfn_extensions/post_hoc_ensembles/sklearn_interface.py
+++ b/src/tabpfn_extensions/post_hoc_ensembles/sklearn_interface.py
@@ -124,7 +124,7 @@ class AutoTabPFNBase(BaseEstimator):
     def _get_predictor_init_args(self) -> dict[str, Any]:
         """Constructs the initialization arguments for AutoGluon's TabularPredictor."""
         timestamp = datetime.datetime.now().strftime("%Y%m%d_%H%M%S")
-        default_args = {"verbosity": 2, "path": f"TabPFNModels/m-{timestamp}"}
+        default_args = {"verbosity": 1, "path": f"TabPFNModels/m-{timestamp}"}
         user_args = self.phe_init_args or {}
         return {**default_args, **user_args}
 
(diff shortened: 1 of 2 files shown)


src/tabpfn_extensions/post_hoc_ensembles/sklearn_interface.py
```diff
@@ -49,7 +49,7 @@ class AutoTabPFNBase(BaseEstimator):
         The evaluation metric for AutoGluon to optimize. If `None`, a default metric
         is chosen based on the problem type (e.g., 'accuracy' for classification).
         For a full list of options, see the AutoGluon documentation.
-    presets : list[str] | str | None, default=None
+    presets : list[str] | str | None, default="best_quality"
```
LeoGrin, line 52: does using "best_quality" default match the max_time default? maybe we should also increase the max_time default because probably 6 min is too small to get nice performance?  Lennart told me AutoGluon ended up using a 1h default time limit and explicitly warn the user that they should define this themselves. Could also be done in aother PR but I was wondering about the interaction with the preset

src/tabpfn_extensions/post_hoc_ensembles/sklearn_interface.py
```diff
@@ -385,12 +386,6 @@ def fit(
     def predict(self, X: pd.DataFrame | np.ndarray) -> np.ndarray:
         check_is_fitted(self)
 
-        if X.shape[1] != self.n_features_in_:
```
LeoGrin, line 388: why was this check faulty acutally?

LeoGrin: ah it because the variable is only set when len(self.classes_) == 1?

LeoGrin: IIUC now this check should be done internally by autogluon correct?

PR #143, opened 2025-08-20
Add TabEBM implementation for data augmentation using SGLD

SilenceX12138:
This pull request introduces the new TabEBM extension to the project, which provides a high-performance method for tabular data augmentation using class-specific energy-based models. The README and project documentation are updated to include TabEBM, and a comprehensive new README for TabEBM is added, detailing its features, installation, and usage.

**Documentation and Extension Integration:**

* Added TabEBM as a new extension for data augmentation in the main project documentation, including a brief description and integration into the module list (`README.md`).
* Updated the project architecture diagram and related sections to include TabEBM as an unsupervised task, along with a direct link to an example notebook (`README.md`). [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R107) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L168-R170) [[3]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R181)

**TabEBM Documentation:**

* Added a comprehensive new `README.md` for TabEBM, describing its purpose, features (such as optimized implementation, GPU acceleration, and memory-efficient sampling), installation instructions, quick start guide, and links to interactive tutorials.

### DIFF
--- a/examples/interpretability/pdp_example.py
+++ b/examples/interpretability/pdp_example.py
@@ -1,0 +1,32 @@
+import matplotlib.pyplot as plt
+from sklearn.datasets import load_breast_cancer
+from sklearn.model_selection import train_test_split
+
+from tabpfn_extensions import TabPFNClassifier
+from tabpfn_extensions.interpretability.pdp import partial_dependence_plots
+
+# Load example dataset
+data = load_breast_cancer()
+X, y = data.data, data.target
+feature_names = list(data.feature_names)
+
+# Split data
+X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.5)
+
+# Initialize and train model
+clf = TabPFNClassifier()
+clf.fit(X_train, y_train)
+
+# 1D PD for the first 3 features + a 2D interaction plot
+disp = partial_dependence_plots(
+    estimator=clf,
+    X=X_test,
+    features=[0, 1, 2, (0, 3)],
+    grid_resolution=30,
+    kind="average",
+    target_class=1,
+)
+disp.figure_.suptitle("Partial dependence")
+
+plt.savefig("pdp_plot.png")
+plt.show()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,6 +11,7 @@ dependencies = [
     "scikit-learn>=1.2.0,<1.7",
     "scipy>=1.11.1,<2",
     "tabpfn>=2.1.1,<3",
+    "tabpfn-common-utils[telemetry-interactive]>=0.2.0",
 ]
 
 requires-python = ">=3.9"
--- a/scripts/generate_dependencies.py
+++ b/scripts/generate_dependencies.py
@@ -14,7 +14,9 @@ from pathlib import Path
 
 def parse_dependency_lines(content: str) -> list[str]:
     """Finds and cleans the dependency lines from the pyproject.toml content."""
-    deps_match = re.search(r"dependencies\s*=\s*\[(.*?)\]", content, re.DOTALL)
+    # Find the dependencies section and match until we find the closing bracket
+    # that's not part of a package extra specification
+    deps_match = re.search(r"dependencies\s*=\s*\[(.*?)\n\]", content, re.DOTALL)
     if not deps_match:
         return []
 
@@ -62,20 +64,21 @@ def main() -> None:
     if args.mode == "maximum":
         for dep in deps:
             # Check for maximum version constraint
-            max_version_match = re.search(r'([^>=<\s]+).*?<\s*([^,\s"\']+)', dep)
+            pattern = r'([^>=<\s]+(?:\[[^\]]+\])?).*?<\s*([^,\s"\']+)'
+            max_version_match = re.search(pattern, dep)
             if max_version_match:
                 package, max_ver = max_version_match.groups()
                 output_reqs.append(f"{package}<{max_ver}")
             else:
                 # If no max version, just use the package name
-                package_match = re.match(r"([^>=<\s]+)", dep)
+                package_match = re.match(r"([^>=<\s]+(?:\[[^\]]+\])?)", dep)
                 if package_match:
                     output_reqs.append(package_match.group(1))
 
     elif args.mode == "minimum":
         for dep in deps:
             # Check for minimum version constraint
-            match = re.match(r'([^>=<\s]+)\s*>=\s*([^,\s"\']+)', dep)
+            match = re.match(r'([^>=<\s]+(?:\[[^\]]+\])?)\s*>=\s*([^,\s"\']+)', dep)
             if match:
                 package, min_ver = match.groups()
                 output_reqs.append(f"{package}=={min_ver}")
--- a/src/tabpfn_extensions/__init__.py
+++ b/src/tabpfn_extensions/__init__.py
@@ -6,6 +6,9 @@ except PackageNotFoundError:
     __version__ = "0.1.0.dev0"
 
 # Create alias for test_utils
+# Import third party extensions
+from tabpfn_common_utils.telemetry.interactive import opt_in
+
 from . import test_utils
 from .embedding import TabPFNEmbedding
 from .hpo import TunedTabPFNClassifier, TunedTabPFNRegressor
@@ -29,3 +32,6 @@ __all__ = [
     "TunedTabPFNClassifier",
     "TunedTabPFNRegressor",
 ]
+
+# Prompt the user to opt in for telemetry and subscribe to the newsletter
+opt_in()
--- a/src/tabpfn_extensions/benchmarking/experiment.py
+++ b/src/tabpfn_extensions/benchmarking/experiment.py
@@ -3,8 +3,10 @@
 from __future__ import annotations
 
 import matplotlib.pyplot as plt
+from tabpfn_common_utils.telemetry import set_extension
 
 
+@set_extension("benchmarking")
 class Experiment:
     """Base class for experiments.
     Experiments should be reproducible, i.e. the settings should give all the information
--- a/src/tabpfn_extensions/embedding/tabpfn_embedding.py
+++ b/src/tabpfn_extensions/embedding/tabpfn_embedding.py
@@ -1,10 +1,12 @@
 from __future__ import annotations
 
 import numpy as np
+from tabpfn_common_utils.telemetry import set_extension
 
 from tabpfn_extensions.utils import TabPFNClassifier, TabPFNRegressor
 
 
+@set_extension("embedding")
 class TabPFNEmbedding:
     """TabPFNEmbedding is a utility for extracting embeddings from TabPFNClassifier or TabPFNRegressor models.
     It supports standard training (vanilla embedding) as well as K-fold cross-validation for embedding extraction.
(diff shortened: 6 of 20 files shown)


SilenceX12138:
> Hi @SilenceX12138, thank you so much this extensive PR, we would love to have this in the package. It appears that right now some of your new test cases are failing. Once these are passing, ping me again and I will review the entire PR :)

Hi @klemens-floege. I have updated the code format, and it passes all the `ruff` checks on my local machine. Could you please check out the latest version when it works? Thank you!

SilenceX12138:
Hi @klemens-floege. I have resolved the compatibility issues related to NumPy’s behaviour across versions and platforms. When you have some time, could you please test the latest changes and let me know if everything works on your end? Thanks!

examples/tabebm/tabebm_augment_real_world_data.ipynb
klemens-floege, line 1: I like this Notebook and got it to run on my end. I have 2 comments: 

- While Notebooks are great for showcasing your methods, they are not ideal for being checked into git + maintained. Since the code does not have visualizations, I think it would be best if we move this code into a single example.py file. 
-You are using tabcamel to load the example data, while this library is quite neat, it seems a bit unnecessary to add a dependency for this. Would it be possible to just use fetch_openml, as in the large_datasets_example.py file. The adult dataset should also be available under ID 45068 I believe.

src/tabpfn_extensions/tabebm/tabebm.py
```diff
@@ -0,0 +1,733 @@
...
+        Args:
+            X: Input features tensor
+            y: Input labels tensor
+
+        Returns:
+            Dictionary containing prepared batch data
+        """
+        # Use efficient train/test split with no validation data
+        splitter = partial(
+            TabEBM.train_test_split_allow_full_train,
```
klemens-floege, line 435: I you want to use all data for train can you not just skip the call entirely? In there case you can use sklearns train_test_split

src/tabpfn_extensions/tabebm/tabebm.py
```diff
@@ -0,0 +1,733 @@
...
+        X: np.ndarray | torch.Tensor | pd.DataFrame,
+        y: np.ndarray | torch.Tensor | pd.Series,
+        num_samples: int,
+        starting_point_noise_std: float = 0.01,
+        sgld_step_size: float = 0.1,
+        sgld_noise_std: float = 0.01,
+        sgld_steps: int = 200,
+        distance_negative_class: float = 5,
+        seed: int = 42,
+        debug: bool = False,
```
klemens-floege: It's better not to have these debug flags lying around. In case you would like to support a setting where we print things during sampling etc. rename to verbose

src/tabpfn_extensions/tabebm/tabebm.py
```diff
@@ -0,0 +1,733 @@
...
+        X: np.ndarray,
+        y: np.ndarray,
+        num_samples: int,
+        starting_point_noise_std: float = 0.01,
+        sgld_step_size: float = 0.1,
+        sgld_noise_std: float = 0.01,
+        sgld_steps: int = 200,
+        distance_negative_class: float = 5,
+        seed: int = 42,
+        debug: bool = False,
```
klemens-floege: Here I would remove debug since either just skips this sampling step, in case we ever want to skip this then lets have another parameter that explicitly states this

src/tabpfn_extensions/tabebm/tabebm.py
```diff
@@ -0,0 +1,733 @@
...
+
+        # Pre-allocate noise tensor for all SGLD steps (memory optimization)
+        sample_shape = (num_samples, X.shape[1])
+        noise_shape = (sgld_steps, *sample_shape)
+
+        # Set seed for reproducible noise generation
+        seed_everything(seed)
+
+        for target_class in unique_classes:
+            if debug:
```
klemens-floege: This is already done above, but as per comments above should probs remove these

src/tabpfn_extensions/tabebm/tabebm.py
```diff
@@ -0,0 +1,733 @@
...
+        for t in range(sgld_steps):
+            # Clear previous gradients
+            if X_sgld_list[0].grad is not None:
+                X_sgld_list[0].grad.zero_()
+
+            # Forward pass to compute energy
+            logits = self.model.forward(X_sgld_list, return_logits=True)
+
+            # Compute energy and total loss
+            energy = TabEBM.compute_energy(logits.reshape(logits.shape[-1], -1))
```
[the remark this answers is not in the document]
klemens-floege: I tested this on my end and got exactly the same results as before @SilenceX12138 what do you think?

```
Vanilla model's balanced accuracy: 71.33
Augmented model's balanced accuracy: 75.99
```

examples/tabebm/tabebm_augment_real_world_data.ipynb
klemens-floege: On a slightly different note I generated 100 samples and got even better results which is great. 

Augmented model's balanced accuracy: 75.99 -> Augmented model's balanced accuracy: 77.60

src/tabpfn_extensions/tabebm/README.md
```diff
@@ -0,0 +1,57 @@
...
+[![Test In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/andreimargeloiu/TabEBM/blob/main/tutorials/tutorial3_augment_real_world_data.ipynb)
+[![Python 3.10+](https://img.shields.io/badge/python-3.10+-blue.svg)](https://www.python.org/downloads/release/python-3100/)
+[![Downloads](https://static.pepy.tech/badge/tabebm)](https://pypi.org/project/tabebm/)
+
+</div>
+
+Official code for the paper ["TabEBM: A Tabular Data Augmentation Method with Distinct Class-Specific Energy-Based Models"](https://arxiv.org/abs/2409.16118), published in the Thirty-Eighth Annual Conference on Neural Information Processing Systems (NeurIPS 2024).
+
+Authored by [Andrei Margeloiu\*](https://www.linkedin.com/in/andreimargeloiu/), [Xiangjian Jiang\*](https://silencex12138.github.io/), [Nikola Simidjievski](https://simidjievskin.github.io/), [Mateja Jamnik](https://www.cl.cam.ac.uk/~mj201/), University of Cambridge, UK
+
```
klemens-floege, line 21: Would be great to list you as primary maintainer as well

```suggestion

**Primary Maintainer:** Xiangjian Jiang

```

src/tabpfn_extensions/tabebm/tabebm.py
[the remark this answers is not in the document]
SilenceX12138: Hi @klemens-floege. Thank you for the review! However, there are a few key points where the current suggestion appears to be based on an incorrect assumption:

**1. Incorrect Assumption About Tensor Shape**
The comment assumes that the shape of logits is `(1, num_samples, num_classes)`, but in fact, the correct shape is `(1, num_classes, num_samples)`. This distinction is critical because reshaping from `(1, C, N)` to `(N, C)` using `logits.reshape(logits.shape[-1], -1)` yields the intended (num_samples, num_classes) shape expected by compute_energy.

**2. Misdiagnosis Based on Static Analysis**
This feedback appears to rely solely on static shape analysis, likely inferred textually or by an LLM. While helpful in many cases, such methods are prone to subtle misinterpretations, especially when tensor semantics are involved. I would recommend verifying tensor shapes at runtime using breakpoints or print statements to avoid incorrect assumptions. :)))

**3. Behaviour Appears Correct Due to Binary Surrogate Setup**
The results may appear consistent because TabEBM currently applies a binary surrogate task, which masks shape-related inconsistencies (i.e., the logit dimension is always 1, leading to equivalent `reshape`/`permute`/`squeeze`). However, the current implementation is deliberately structured to support more general settings, and any changes here could break behaviour in multiclass scenarios or under structural assumptions used elsewhere in the model.

For these reasons, I suggest retaining the existing reshaping logic as it correctly aligns with the input expectations of `compute_energy` and supports the broader design of the model.

src/tabpfn_extensions/tabebm/tabebm.py
SilenceX12138: 1. You might consider trying `train_test_split`, though it appears not to support a full training setup (i.e., using the entire dataset with `test_size=0`).
2. Additionally, the `split_fn` argument is mandatory for TabPFN’s `get_preprocessed_datasets`. Skipping it would require more fundamental changes to the TabPFN model itself. If you're interested in modifying it, I’d be glad to see TabPFN become more flexible in this regard.

Given these constraints, I’d recommend keeping the current code unchanged.

src/tabpfn_extensions/tabebm/tabebm.py
SilenceX12138: Renamed

examples/tabebm/tabebm_augment_real_world_data.ipynb
SilenceX12138: Added a script as requested

SilenceX12138:
Hi @klemens-floege, thanks for flagging this. I’ve double-checked on my local machine, and the results seem consistent there. The difference you’re seeing is likely due to hardware variances rather than an implementation issue. On my end, I can robustly reproduce the results shown below.

It’s also worth noting that the method is not expected to consistently outperform the original data in every run. The improvements reported in the paper are based on averages across multiple splits, rather than on a single run.

In short, I’d suggest keeping the code as it is.


```shell
Loading adult dataset from OpenML...
Dataset shape: (48842, 14)
Target classes: ['<=50K' '>50K']
Feature names: ['age', 'fnlwgt', 'education-num', 'capital-gain', 'capital-loss', 'hours-per-week', 'workclass', 'education', 'marital-status', 'occupation', 'relationship', 'race', 'sex', 'native-country']
Subsampling dataset to 200 samples...
Subsampled dataset shape: (200, 14)
Class distribution: class
<=50K    152
>50K      48
Name: count, dtype: int64
Training set size: 160
Test set size: 40
Categorical features: ['workclass', 'education', 'marital-status', 'occupation', 'relationship', 'race', 'sex', 'native-country']
Numerical features: ['age', 'fnlwgt', 'education-num', 'capital-gain', 'capital-loss', 'hours-per-week']
Processed training data shape: (160, 68)
Processed test data shape: (40, 68)
Unique target labels: [0 1]

==================================================
GENERATING SYNTHETIC DATA WITH TabEBM
==================================================
Original training set size: 160
Synthetic data size: 100
Augmented training set size: 260

==================================================
TRAINING DOWNSTREAM PREDICTORS
==================================================

Training vanilla model on original data...
Training augmented model on real + synthetic data...

==================================================
EVALUATION RESULTS
==================================================
Vanilla model's balanced accuracy: 66.67%
Augmented model's balanced accuracy: 68.33%
Improvement: 1.67 percentage points
✓ Data augmentation with TabEBM improved model performance!
```

PR #148, opened 2025-08-26
AutoTabPFN Default Presets to None

klemens-floege:


examples/phe/phe_example.py
```diff
@@ -4,6 +4,9 @@
 """WARNING: This example may run slowly on CPU-only systems.
 For better performance, we recommend running with GPU acceleration.
 This example trains multiple TabPFN models, which is computationally intensive.
+
+NOTE: For higher accuracy, consider manually setting 'presets="best_quality'
```
noahho: Can we instead have an inline comment:

```
clf = AutoTabPFNClassifier(
max_time=3600,
# Adjust the preset to a lower setting, e.g. "medium" when less time is available
# When running "best_quality" with insufficient time, the run will error with no models fitted
preset="best_quality"
) 
```

noahho (changes requested):
Looks good. For the preset I'd however add the comment right next to the parameter

noahho (approved):

PR #152, opened 2025-08-29
finetuning wrapper

anuragg1209:
Hi @noahho & @klemens-floege,

Below is the code for the fine-tuning wrapper. Could you please review it when you get a chance? I’d appreciate any suggestions for improvements or changes to make it better. Thanks a lot!

### DIFF
--- a/examples/finetune/finetune_example.py
+++ b/examples/finetune/finetune_example.py
@@ -1,0 +1,62 @@
+from sklearn.datasets import fetch_covtype
+from sklearn.model_selection import train_test_split
+from sklearn.metrics import log_loss, roc_auc_score
+import numpy as np
+import torch
+from tabpfn import TabPFNClassifier
+
+from tabpfn_extensions.finetune.finetune_classifier import FinetunedTabPFNClassifier
+
+# 1. Load and prepare the data
+# We use a small subset for a quick demonstration.
+print("--- 1. Loading Data ---")
+X_all, y_all = fetch_covtype(return_X_y=True, shuffle=True)
+X, y = X_all[:10000], y_all[:10000]
+
+# Create a final hold-out test set. This is NOT used during fine-tuning.
+X_train, X_test, y_train, y_test = train_test_split(
+    X, y, test_size=0.2, random_state=42, stratify=y
+)
+print(f"Data split: {len(X_train)} training samples, {len(X_test)} test samples.\n")
+
+
+# 2. Initial model evaluation on test set
+
+base_clf = TabPFNClassifier(device='cuda' if torch.cuda.is_available() else 'cpu', n_estimators=2)
+base_clf.fit(X_train, y_train)
+
+roc_auc = roc_auc_score(y_test, base_clf.predict_proba(X_test)[:, 1]) if len(np.unique(y_test)) == 2 else roc_auc_score(y_test, base_clf.predict_proba(X_test), multi_class="ovr", average="weighted")
+log_loss_score = log_loss(y_test, base_clf.predict_proba(X_test))
+
+print(f"📊 Initial Test ROC: {roc_auc:.4f}")
+print(f"📊 Initial Test Log Loss: {log_loss_score:.4f}\n")
+
+# 2. Initialize and run fine-tuning
+print("--- 2. Initializing and Fitting Model ---\n")
+
+# Instantiate the wrapper with your desired hyperparameters
+finetuned_clf = FinetunedTabPFNClassifier(
+    device='cuda' if torch.cuda.is_available() else 'cpu',
+    epochs=10,
+    learning_rate=1e-5,
+    meta_dataset_size=1024,
+    finetune_split_ratio=0.3,
+    random_state=42,
+    n_estimators=2,
+    patience=3
+)
+
+# Call .fit() to start the fine-tuning process on the training data
+finetuned_clf.fit(X_train, y_train)
+print("\n")
+
+# 3. Evaluate the fine-tuned model
+print("--- 3. Evaluating Model on Held-out Test Set ---\n")
+y_pred_proba = finetuned_clf.predict_proba(X_test)
+y_pred = finetuned_clf.predict(X_test)
+
+roc_auc = roc_auc_score(y_test, y_pred_proba[:, 1]) if len(np.unique(y_test)) == 2 else roc_auc_score(y_test, y_pred_proba, multi_class="ovr", average="weighted")
+loss = log_loss(y_test, y_pred_proba)
+
+print(f"📊 Final Test ROC: {roc_auc:.4f}")
+print(f"📊 Final Test Log Loss: {loss:.4f}")
--- a/src/tabpfn_extensions/finetune/finetune_classifier.py
+++ b/src/tabpfn_extensions/finetune/finetune_classifier.py
@@ -1,0 +1,308 @@
+import torch
+from torch.optim import Adam
+from torch.utils.data import DataLoader
+from sklearn.base import BaseEstimator, ClassifierMixin
+from sklearn.model_selection import train_test_split
+from sklearn.metrics import roc_auc_score, log_loss
+from sklearn.utils.validation import check_is_fitted
+from functools import partial
+from tqdm.auto import tqdm
+import numpy as np
+
+from tabpfn import TabPFNClassifier
+from tabpfn.finetune_utils import clone_model_for_evaluation
+from tabpfn.utils import meta_dataset_collator
+
+
+def evaluate_model(
+        classifier: TabPFNClassifier,
+        eval_config: dict,
+        X_train: np.ndarray,
+        y_train: np.ndarray,
+        X_val: np.ndarray,
+        y_val: np.ndarray,
+    ) -> tuple[float, float]:
+        """Evaluates the model's performance on the validation set."""
+        eval_classifier = clone_model_for_evaluation(
+            classifier, eval_config, TabPFNClassifier
+        )
+        eval_classifier.fit(X_train, y_train)
+
+        try:
+            probabilities = eval_classifier.predict_proba(X_val) # type: ignore
+            roc_auc = roc_auc_score(
+                y_val, probabilities, multi_class="ovr", average="weighted"
+            ) if len(np.unique(y_val)) > 2 else roc_auc_score(
+                y_val, probabilities[:, 1]
+            )
+            log_loss_score = log_loss(y_val, probabilities)
+        except Exception as e:
+            print(f"An error occurred during evaluation: {e}")
+            roc_auc, log_loss_score = np.nan, np.nan
+
+        return roc_auc, log_loss_score
+
+
+class FinetunedTabPFNClassifier(BaseEstimator, ClassifierMixin):
+    """
+    A scikit-learn compatible wrapper for fine-tuning the TabPFNClassifier.
+
+    This class encapsulates the fine-tuning loop, allowing you to fine-tune
+    TabPFN on a specific dataset using the familiar .fit() and .predict() API.
+
+    Parameters
+    ----------
+    device : str, default='cuda'
+        The device to run the model on, e.g., 'cuda' or 'cpu'.
+
+    epochs : int, default=5
+        The total number of passes through the fine-tuning data.
+
+    learning_rate : float, default=1e-5
+        The learning rate for the Adam optimizer. A small value is crucial
+        for stable fine-tuning.
+
+    meta_dataset_size : int, default=1024
+        The total number of samples to use for creating each meta-dataset, which
+        is then split into a context set and a fine-tuning set. This should be
+        less than or equal to the number of samples in the training data.
+
+    finetune_split_ratio : float, default=0.3
+        The proportion of each meta-dataset to use for calculating the loss.
+        The rest is used as the context for the model.
+
+    meta_batch_size : int, default=1
+        The number of meta-datasets to process in a single optimization step.
+        Currently, this should be kept at 1.
+
+    random_state : int, default=42
+        Seed for reproducibility of data splitting and model initialization.
+
+    early_stopping : bool, default=True
+        Whether to use early stopping based on validation ROC AUC performance.
+
+    patience : int, default=3
+        Number of epochs to wait for improvement before early stopping.
+
+    min_delta : float, default=1e-4
+        Minimum change in ROC AUC to be considered as an improvement.
+
+    **kwargs : dict
+        Additional keyword arguments to pass to the underlying TabPFNClassifier,
+        such as `n_estimators`.
+    """
+    def __init__(self,
+                 device: str = 'cuda',
+                 epochs: int = 5,
+                 learning_rate: float = 1e-5,
+                 meta_dataset_size: int = 1024,
+                 finetune_split_ratio: float = 0.3,
+                 meta_batch_size: int = 1,
+                 random_state: int = 42,
+                 early_stopping: bool = True,
+                 patience: int = 3,
+                 min_delta: float = 1e-4,
+                 **kwargs): # type: ignore
+        super().__init__()
+        self.device = device
+        self.epochs = epochs
+        self.learning_rate = learning_rate
+        self.meta_dataset_size = meta_dataset_size
+        self.finetune_split_ratio = finetune_split_ratio
+        self.meta_batch_size = meta_batch_size
+        self.random_state = random_state
+        self.early_stopping = early_stopping
+        self.patience = patience
+        self.min_delta = min_delta
+        self.kwargs = kwargs
+
+    
+    def fit(self, X: np.ndarray, y: np.ndarray) -> "FinetunedTabPFNClassifier":
+        """
+        Fine-tunes the TabPFN model on the provided training data.
+
+        Parameters
+        ----------
+        X : array-like of shape (n_samples, n_features)
+            The training input samples.
+
+        y : array-like of shape (n_samples,)
+            The target values.
+
+        Returns
+        -------
+        self : object
+            Returns the instance itself.
+        """
+    
+        self.X_ = X
+        self.y_ = y
+
+        classifier_config = {
+        "ignore_pretraining_limits": True,
+        "device": self.device,
+        "n_estimators": self.kwargs.get("n_estimators", 8),
+        "random_state": self.random_state,
+        "inference_precision": torch.float32,
+        }
+        
+
+        # 2. Initialize the base TabPFNClassifier
+        self.classifier_ = TabPFNClassifier(**classifier_config, fit_mode="batched", differentiable_input=False)
+        # Required to access model parameters for the optimizer
+        self.classifier_._initialize_model_variables()
+
+        eval_config = {
+            **classifier_config,
+             "inference_config": {
+                "SUBSAMPLE_SAMPLES": self.kwargs.get("n_inference_context_samples", 10000)
+            }
+        }
+
+        # 3. Prepare data for the fine-tuning loop
+        # This splitter function will be applied to the training data to create
+        # (context, finetune) pairs for each step of the loop.
+        splitter = partial(
+            train_test_split,
+            test_size=self.finetune_split_ratio,
+            random_state=self.random_state
+        )
+
+        X_train, X_val, y_train, y_val = splitter(X, y)
+
+        # Cap the meta-dataset size by the total number of training samples
+        effective_meta_dataset_size = min(self.meta_dataset_size, len(y_train))
+
+        training_datasets = self.classifier_.get_preprocessed_datasets(
+            X_train, y_train, splitter, effective_meta_dataset_size
+        )
+
+        finetuning_dataloader = DataLoader(
+            training_datasets,
+            batch_size=self.meta_batch_size,
+            collate_fn=meta_dataset_collator,
+            shuffle=True
+        )
+
+        # 4. Setup optimizer and loss function
+        optimizer = Adam(self.classifier_.model_.parameters(), lr=self.learning_rate)  # type: ignore
+        loss_function = torch.nn.CrossEntropyLoss()
+
+
+        # 5. Fine-tuning loop
+        print("--- 🚀 Starting Fine-tuning ---")
+        
+        # Early stopping variables
+        best_roc_auc = -np.inf
+        patience_counter = 0
+        best_model_state = None
+        
+        for epoch in range(self.epochs):
+            progress_bar = tqdm(
+                finetuning_dataloader,
+                desc=f"Finetuning Epoch {epoch + 1}/{self.epochs}"
+            )
+            for (
+                X_context_batch, # Context features
+                X_query_batch,  # Query features
+                y_context_batch, # Context labels
+                y_query_batch,  # Query labels
+                cat_ixs,
+                confs,
+            ) in progress_bar:
+
+                # Skip batches where a split results in missing classes,
+                # which can cause errors in loss calculation.
+                if len(np.unique(y_context_batch)) != len(np.unique(y_query_batch)):
+                    continue
+
+                optimizer.zero_grad()
+
+                # Provide the context set to the model
+                self.classifier_.fit_from_preprocessed(
+                    X_context_batch, y_context_batch, cat_ixs, confs
+                )
+
+                # Get predictions (logits) on the fine-tuning set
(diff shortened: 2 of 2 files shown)


anuragg1209:
> Thanks for the PR @anuragg1209, when looking at the logs that you shared, it seems like the fine-tuning is not really helping. Shall we still merge it though and do parameter tweaking later or should we figure this out first?

Hi Klemens, it would be helpful to first review the code and ensure there's no bug in the wrapper that's preventing the fine-tuning from being effective, or if it's just the hyperparameter settings that need optimization. A thorough review might be beneficial before merging the code.

examples/finetune/finetune_example.py
```diff
@@ -0,0 +1,63 @@
...
+)
+
+# Call .fit() to start the fine-tuning process on the training data
+finetuned_clf.fit(X_train, y_train)
+print("\n")
+
+# 3. Evaluate the fine-tuned model
+print("--- 3. Evaluating Model on Held-out Test Set ---\n")
+y_pred_proba = finetuned_clf.predict_proba(X_test)
+y_pred = finetuned_clf.predict(X_test)
```
noahho:
```suggestion
```

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,308 @@
...
+        try:
+            probabilities = eval_classifier.predict_proba(X_val) # type: ignore
+            roc_auc = roc_auc_score(
+                y_val, probabilities, multi_class="ovr", average="weighted"
+            ) if len(np.unique(y_val)) > 2 else roc_auc_score(
+                y_val, probabilities[:, 1]
+            )
+            log_loss_score = log_loss(y_val, probabilities)
+        except Exception as e:
+            print(f"An error occurred during evaluation: {e}")
```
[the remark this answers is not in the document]
noahho: +1

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        # 3. Prepare data for the fine-tuning loop
+        # This splitter function will be applied to the training data to create
+        # (context, finetune) pairs for each step of the loop.
+        splitter = partial(
+            train_test_split,
+            test_size=self.finetune_split_ratio,
+            random_state=self.random_state
+        )
```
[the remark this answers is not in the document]
noahho: +1

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,308 @@
...
+                f"📊 Epoch {epoch + 1} Evaluation | Val ROC: {roc_auc:.4f}, Val Log Loss: {log_loss_score:.4f}\n"
+            )
+            
+            # Early stopping logic
+            if self.early_stopping and not np.isnan(roc_auc):
+                if roc_auc > best_roc_auc + self.min_delta:
+                    best_roc_auc = roc_auc
+                    patience_counter = 0
+                    # Save the best model state
+                    best_model_state = {key: value.clone() for key, value in self.classifier_.model_.state_dict().items()}  # type: ignore
```
noahho: this might be prone to go wrong, not fully loading the checkpoint. Instead we could save and reload models using the tabpfns public functions: https://github.com/PriorLabs/TabPFN/blob/main/src/tabpfn/model_loading.py#L590

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,308 @@
...
+        -------
+        p : array of shape (n_samples, n_classes)
+            The class probabilities of the input samples.
+        """
+        check_is_fitted(self)
+
+        # For prediction, TabPFN needs to see the full training data as context
+        # Set fit_mode to standard mode for prediction to avoid the warning
+        self.classifier_.fit_mode = "fit_preprocessors"
+        self.classifier_.fit(self.X_, self.y_)
```
klemens-floege: I think should be in the fit() part, this way we can use it straight away

noahho: Move this to the fit method (at the end of it)

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,308 @@
...
+        "ignore_pretraining_limits": True,
+        "device": self.device,
+        "n_estimators": self.kwargs.get("n_estimators", 8),
+        "random_state": self.random_state,
+        "inference_precision": torch.float32,
+        }
+        
+
+        # 2. Initialize the base TabPFNClassifier
+        self.classifier_ = TabPFNClassifier(**classifier_config, fit_mode="batched", differentiable_input=False)
```
noahho: Can we rename classifier to finetuned_classifier_

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,308 @@
...
+        -------
+        y_pred : array of shape (n_samples,)
+            The predicted classes.
+        """
+        check_is_fitted(self)
+
+        # For prediction, TabPFN needs to see the full training data as context
+        # Set fit_mode to standard mode for prediction to avoid the warning
+        self.classifier_.fit_mode = "fit_preprocessors"
+        self.classifier_.fit(self.X_, self.y_)
```
klemens-floege: sam as above

src/tabpfn_extensions/finetune/finetune_classifier.py
klemens-floege: Add stratify=y here

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,308 @@
...
+                # which can cause errors in loss calculation.
+                if len(np.unique(y_context_batch)) != len(np.unique(y_query_batch)):
+                    continue
+
+                optimizer.zero_grad()
+
+                # Provide the context set to the model
+                self.classifier_.fit_from_preprocessed(
+                    X_context_batch, y_context_batch, cat_ixs, confs
+                )
```
noahho: Could you check (and log) whats the size of the context batches? I think could be a problem is the context sizes. This might be the reason why final test doesn't work well:
1.) Always having shape context size means we might not generalize to full_training set size
2.) If we have 9K samples, 4K context sizes, we get batches: [4K, 4K, 1K] This means we are predicting on some awkwardly shaped batches. We should drop the non-filled batches, I started on this here: https://github.com/PriorLabs/TabPFN/pulls

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,308 @@
...
+        "inference_precision": torch.float32,
+        }
+        
+
+        # 2. Initialize the base TabPFNClassifier
+        self.classifier_ = TabPFNClassifier(**classifier_config, fit_mode="batched", differentiable_input=False)
+        # Required to access model parameters for the optimizer
+        self.classifier_._initialize_model_variables()
+
+        eval_config = {
```
noahho, line 208: The variable "inference_config" -> "SUBSAMPLE_SAMPLES" is only set in the valid loops but not finally after training? We could rather set this in the classifier_config which is given to the final classifier too?

noahho: The size of the chunks in finetuning should be variable or exactly match what the model gets in final application. I.e. SUBSAMPLE_SAMPLES == context_size == final classifier SUBSAMPLE_SAMPLES

noahho (changes requested):

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,308 @@
...
+                X_query_batch,  # Query features
+                y_context_batch, # Context labels
+                y_query_batch,  # Query labels
+                cat_ixs,
+                confs,
+            ) in progress_bar:
+
+                # Skip batches where a split results in missing classes,
+                # which can cause errors in loss calculation.
+                if len(np.unique(y_context_batch)) != len(np.unique(y_query_batch)):
```
klemens-floege: We could have something like this. However skipping might also be smarter, however then lets 

A: log that 
B: at the end print like "so and so many out of total" batches were skipped each epoch. :

```
ctx = set(np.unique(y_context_batch))
qry = set(np.unique(y_query_batch))
if not qry.issubset(ctx):
    continue
```

src/tabpfn_extensions/finetune/finetune_classifier.py
anuragg1209: Fixed! Now we save and reload model using tabpfns public functions.

src/tabpfn_extensions/finetune/finetune_classifier.py
anuragg1209: Resolved.

src/tabpfn_extensions/finetune/finetune_classifier.py
anuragg1209: Resolved. Now we have `finetuned_classifier_` instead of just `classifier_ ` .

src/tabpfn_extensions/finetune/finetune_classifier.py
anuragg1209: Yes, after debugging, we noticed the batches were not filled during fine-tuning. This issue has been fixed with this PR in the main `tabpfn` repo: 
https://github.com/PriorLabs/TabPFN/commit/648e3d836b9b74f19ad3cf89e9be670c5a388563

So now, if we have 9k samples, we get a single batch of full 9k samples instead of [4k, 4k, 1k] as done previously. 
If there are supposed to be more than 10k samples, eg, 24k, then we will have [10k, 10k, 4k] batches. And in the finetuning loop, only filled batches are processed and non-filled batches skipped.

src/tabpfn_extensions/finetune/finetune_classifier.py
anuragg1209: > The variable "inference_config" -> "SUBSAMPLE_SAMPLES" is only set in the valid loops but not finally after training? We could rather set this in the classifier_config which is given to the final classifier too?

Resolved. Now we have the same `inference_config` for the final classifier. But we still need to create `eval_config` separately from the `classifier_config` as `"inference_config" -> "SUBSAMPLE_SAMPLES"` in the `eval_config` is not accepted by the Dataloader during finetuning, which results in other unexpected errors.

anuragg1209: > The size of the chunks in finetuning should be variable or exactly match what the model gets in final application. I.e. SUBSAMPLE_SAMPLES == context_size == final classifier SUBSAMPLE_SAMPLES

Fixed. Now the `context _size == SUBSAMPLE_SAMPLES == final classifier SUBSAMPLE_SAMPLES `. The model now sees the same number of samples during the finetuning and during inference.

src/tabpfn_extensions/finetune/finetune_classifier.py
anuragg1209: Fixed. Moved the statements to the fit() part.

src/tabpfn_extensions/finetune/finetune_classifier.py
anuragg1209: Fixed.

src/tabpfn_extensions/finetune/finetune_classifier.py
anuragg1209: Fixed. Now we have a different splitter for validation data and training data, and it has a `stratify=y` parameter.

src/tabpfn_extensions/finetune/finetune_classifier.py
anuragg1209: We simply skip the batches that do not contain all the classes. The frequency of batches getting skipped will be quite lower, as now we get batches that are filled due to other fixes made above.

noahho: I think it's still a good idea to log this!

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,395 @@
...
+        such as `n_estimators`.
+    """
+
+    def __init__(
+        self,
+        device: str = "cuda",
+        epochs: int = 5,
+        learning_rate: float = 1e-5,
+        n_inference_context_samples: int = 10_000,
+        finetune_split_ratio: float = 0.2,
```
noahho, line 133: I'd rename this to validation ratio. In my understanding this poart of the data is used for validation while finetuning, but finetuning is obviosu given the context
```suggestion
        validation_ratio: float = 0.2,
```

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,395 @@
...
+        }
+
+        # Prepare data for the fine-tuning loop
+        # This splitter function will be applied to the training data to create
+        # (context, query) pairs for each step of the loop.
+
+        training_splitter = partial(
+            train_test_split,
+            test_size=self.finetune_split_ratio,
+            random_state=self.random_state,
```
noahho: This could be a problem and lead to significantly degraded performance?

Static Meta-Dataset Splits: The training_splitter uses a fixed random_state.

The Problem: This means that in every epoch, the model sees the exact same context/query pairs drawn from the training data. This lack of stochasticity can reduce the diversity of the training signal.

Suggestion: For potentially more robust training, the splits could be regenerated with a different seed each epoch. This could be achieved by creating a new partial function for the splitter inside the epoch loop, using a different random_state each time (e.g., self.random_state + epoch).

noahho: Not sure if this is actually randomized already though!

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,395 @@
...
+            stratify=y,
+        )
+
+        X_train, X_val, y_train, y_val = validation_splitter(X, y)
+
+        # Calculate the context size used during finetuning
+        context_size = min(self.n_inference_context_samples, len(y_train))
```
noahho: I think we'd actually want to make this self.n_inference_context_samples/(1-self.finetune_split_ratio). This way we finetune with a trainset of self.n_inference_context_samples right? Or is that already the case? My worry is we are splitting out the valid set fro. this again making it smaller in finetuning vs final inference

noahho:
```suggestion
        n_finetuning_fit_predict_context_samples = min(self.n_inference_context_samples, len(y_train))
```

noahho: Very long name but it would clarify that this is fit and predict and it would make consistent with the otehr context samples

noahho (commented):
I made some minor comments and would also add these small points:
On CUDA, wrap the forward/loss/backward with torch.cuda.amp.autocast() and GradScaler; add gradient clipping (e.g., torch.nn.utils.clip_grad_norm_(..., 1.0)).

Learning Rate Schedulers: Fine-tuning often benefits greatly from a learning rate that changes over time (e.g., linear decay or cosine annealing). You could add a scheduler parameter to __init__ that accepts a PyTorch scheduler instance, and then call scheduler.step() after each epoch.

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,395 @@
...
+                ctx = set(np.unique(y_context_batch))
+                qry = set(np.unique(y_query_batch))
+                if not qry.issubset(ctx):
+                    continue
+
+                if (
+                    X_context_batch[0].shape[1] + X_query_batch[0].shape[1]
+                    != context_size
+                ):
+                    continue
```
noahho, line 288: Can we please log if this happens

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,395 @@
...
+                y_context_batch,  # Context labels
+                y_query_batch,  # Query labels
+                cat_ixs,
+                confs,
+            ) in progress_bar:
+                
+                ctx = set(np.unique(y_context_batch))
+                qry = set(np.unique(y_query_batch))
+                if not qry.issubset(ctx):
+                    continue
```
noahho, line 276: can we log if this happens

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,395 @@
...
+                X_query_batch,  # Query features
+                y_context_batch,  # Context labels
+                y_query_batch,  # Query labels
+                cat_ixs,
+                confs,
+            ) in progress_bar:
+                
+                ctx = set(np.unique(y_context_batch))
+                qry = set(np.unique(y_query_batch))
+                if not qry.issubset(ctx):
```
noahho, line 272: why would this happen, can you add a comment?

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,395 @@
...
+                    continue
+
+                if (
```
noahho, line 278: why would this happen, can you add a comment?

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,395 @@
...
+
+        X_train, X_val, y_train, y_val = validation_splitter(X, y)
+
+        # Calculate the context size used during finetuning
+        context_size = min(self.n_inference_context_samples, len(y_train))
+
+        classifier_config = {
+            "ignore_pretraining_limits": True,
+            "device": self.device,
+            "n_estimators": self.kwargs.get("n_estimators", 8),
```
noahho: can we just unpack kwargs here directly? Would allow us to specify any hyperparameter

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,441 @@
...
+
+        # Setup optimizer and loss function
+        optimizer = Adam(
+            self.finetuned_classifier_.model_.parameters(),  # type: ignore
+            lr=self.learning_rate,
+        )
+        loss_function = torch.nn.CrossEntropyLoss()
+
+        # Initialize learning rate scheduler with cosine annealing
+        scheduler = CosineAnnealingLR(optimizer, T_max=self.epochs)
```
bejaeger, line 221: Can you please go through the entire file and remove the unnecessary comments that just describe the code?
For example here and also above for the optimizer and in lots of other places.

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,441 @@
...
+class FinetunedTabPFNClassifier(BaseEstimator, ClassifierMixin):
+    """A scikit-learn compatible wrapper for fine-tuning the TabPFNClassifier.
+
+    This class encapsulates the fine-tuning loop, allowing you to fine-tune
+    TabPFN on a specific dataset using the familiar .fit() and .predict() API.
+
+    Parameters:
+    ----------
+    device : str, default='cuda'
+        The device to run the model on, e.g., 'cuda' or 'cpu'.
```
bejaeger: I guess you can remove "cpu", I guess nobody should run this on cpu.

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,441 @@
...
+    learning_rate : float, default=1e-5
+        The learning rate for the Adam optimizer. A small value is crucial
+        for stable fine-tuning.
+
+    n_inference_context_samples : int, default=10_000
+        The total number of samples to use for creating each meta-dataset, which
+        is then split into a context set and a query set for each step of the fine-tuning loop.
+        This same context size is used during final inference as well.
+
+    finetune_split_ratio : float, default=0.2
```
bejaeger, line 98: seems this has been renamed to `validation_ratio`. Maybe check for all the arguments.

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,441 @@
...
+        self.y_ = y
+
+        validation_splitter = partial(
+            train_test_split,
+            test_size=self.finetune_split_ratio,
+            random_state=self.random_state,
+            stratify=y,
+        )
+
+        X_train, X_val, y_train, y_val = validation_splitter(X, y)
```
bejaeger: looks like `validation_splitter` isn't re-used. Could you simplify and not create a partial function first but use `train_test_split` in one call?

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,441 @@
...
+        # Unpack kwargs to allow any TabPFNClassifier hyperparameter to be specified,
+        # then override with required config values
+        classifier_config = {
+            **self.kwargs,
+            "ignore_pretraining_limits": True,
+            "device": self.device,
+            "random_state": self.random_state,
+        }
+
+        # Initialize the base TabPFNClassifier
```
bejaeger: Remove comment, see also other comment about comments in general. Let's clean things up :)

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,441 @@
...
+                random_state=self.random_state + epoch,
+            )
+
+            training_datasets = self.finetuned_classifier_.get_preprocessed_datasets(
+                X_train,
+                y_train,
+                training_splitter,
+                n_finetuning_fit_predict_context_samples,
+                equal_split_size=False,
+            )
```
bejaeger, line 249: this is a long list of arguments, let's make them keyword arguments.

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,441 @@
...
+                progress_bar.set_postfix(loss=f"{loss.item():.4f}")
+
+            roc_auc, log_loss_score = evaluate_model(
+                self.finetuned_classifier_,
+                eval_config,
+                X_train,  # pyright: ignore[reportArgumentType]
+                y_train,  # pyright: ignore[reportArgumentType]
+                X_val,    # pyright: ignore[reportArgumentType]
+                y_val,    # pyright: ignore[reportArgumentType]
+            )
```
bejaeger, line 346: same as above, let's make them keyword arguments for easier readability

src/tabpfn_extensions/finetune/finetune_classifier.py
```diff
@@ -0,0 +1,441 @@
...
+                    if best_model is not None:
+                        self.finetuned_classifier_ = best_model
+                    break
+
+        logging.info("--- ✅ Fine-tuning Finished ---")
+
+        finetuned_inference_classifier = clone_model_for_evaluation(
+            self.finetuned_classifier_, # type: ignore
+            eval_config,
+            TabPFNClassifier,
```
bejaeger, line 378: keyword arguments please

examples/finetune/finetune_example.py
```diff
@@ -0,0 +1,74 @@
...
+# We use a small subset for a quick demonstration.
+print("--- 1. Loading Data ---")
+X_all, y_all = fetch_covtype(return_X_y=True, shuffle=True)
+X, y = X_all[:11000], y_all[:11000]
+
+# df = pd.read_csv("/home/anurag_priorlabs_ai/tabpfn-extensions/PrudentialLifeInsuranceAssessment.csv")
+
+# print(df.columns)
+# X = df.drop(columns=["Id", "Response"])
+# y = df["Response"]
```
bejaeger: can you clean up the commented code?

bejaeger (approved):
Very nice, thanks @anuragg1209 ! 
I approved so that you can merge later, but would be great if you could implement the (mostly stylistic) things I commented on.
@klemens-floege @noahho , we decided to get this in sooner rather than later so we can iterate on it more easily; hope that's fine.

anuragg1209:
### PR Paused — Development Relocated

Hi, further development of this feature is continuing in a separate private repository. This PR will remain open in its current state for now and will be updated later.

PR #162, opened 2025-09-12
Set up extension context for usage analytics

safaricd:
## Change & Context

TabPFN `2.1.5` introduced anonymous usage telemetry. The change I propose ads context to each telemetry API call - the name of the extension specifically. 

The `set_extension` function can be applied to **both classes and functions**. In case of the former, we automatically apply the decorator to each public function of a class.

### DIFF
--- a/src/tabpfn_extensions/hpo/tuned_tabpfn.py
+++ b/src/tabpfn_extensions/hpo/tuned_tabpfn.py
@@ -57,6 +57,7 @@ from tabpfn_extensions.scoring.scoring_utils import (
     score_classification,
     score_regression,
 )
+from tabpfn_extensions.utils import DeviceSpecification
 
 # Import TabPFN models from extensions (which handles backend compatibility)
 try:
@@ -103,7 +104,7 @@ class TunedTabPFNBase(BaseEstimator):
         n_validation_size: float = 0.2,
         shuffle_data: bool = True,
         metric: MetricType = MetricType.ACCURACY,
-        device: str = "auto",
+        device: DeviceSpecification = "auto",
         random_state: int | None = None,
         categorical_feature_indices: list[int] | None = None,
         verbose: bool = True,
@@ -212,9 +213,9 @@ class TunedTabPFNBase(BaseEstimator):
             }
             model_params["inference_config"] = inference_config
             # Use device utility for automatic selection
-            from tabpfn_extensions.utils import infer_device_and_type
+            from tabpfn_extensions.utils import infer_device
 
-            model_params["device"] = infer_device_and_type(self.device)
+            model_params["device"] = infer_device(self.device)
             model_params["random_state"] = rng.randint(0, 2**31 - 1)
 
             # Handle model type selection
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -3,6 +3,9 @@ on:
   pull_request:
     branches:
       - main
+  push:
+    branches: [main]
+  workflow_dispatch:
 
 jobs:
   check_python_linting:
@@ -18,7 +21,7 @@ jobs:
         with:
           src: "./src ./tests"
           version: 0.8.6
-          args: 'format --check'
+          args: "format --check"
 
   test_compatibility:
     name: Test Package Compatibility
--- a/examples/interpretability/pdp_example.py
+++ b/examples/interpretability/pdp_example.py
@@ -1,0 +1,32 @@
+import matplotlib.pyplot as plt
+from sklearn.datasets import load_breast_cancer
+from sklearn.model_selection import train_test_split
+
+from tabpfn_extensions import TabPFNClassifier
+from tabpfn_extensions.interpretability.pdp import partial_dependence_plots
+
+# Load example dataset
+data = load_breast_cancer()
+X, y = data.data, data.target
+feature_names = list(data.feature_names)
+
+# Split data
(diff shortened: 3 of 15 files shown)


src/tabpfn_extensions/__init__.py
```diff
@@ -29,3 +32,6 @@
     "TunedTabPFNClassifier",
     "TunedTabPFNRegressor",
 ]
+
+# Opt in to telemetry
+opt_in()
```
noahho: I don't assume we are just opting in by calling this, rather it prompts for opt in?

src/tabpfn_extensions/hpo/tuned_tabpfn.py
```diff
@@ -91,6 +92,7 @@ class MetricType(str, Enum):
     MAE = "mae"
 
 
+@set_extension("hpo")
```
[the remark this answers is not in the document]
noahho: Seems legit

src/tabpfn_extensions/rf_pfn/sklearn_based_random_forest_tabpfn.py
```diff
@@ -90,6 +91,7 @@ def _validate_tabpfn(self):
                     "Please use TabPFNRegressor with RandomForestTabPFNRegressor.",
                 )
 
+    @set_extension("rf_pfn")
```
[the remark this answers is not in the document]
noahho: This seems right

src/tabpfn_extensions/unsupervised/unsupervised.py
```diff
@@ -59,6 +60,7 @@
 )
 
 
+@set_extension("unsupervised")
```
noahho: Unsupervised is highly heterogeneous. We should mark the cpaabilities below it such as outlier detection, generation, imputation (while imputation is used internall and as public api)

noahho: + embeddings

safaricd: @noahho I have literally mapped the Python package name -> extension name. This naming is arbitrary and we can change it to whatever makes most sense 😊

src/tabpfn_extensions/__init__.py
safaricd: True, we can just changing the wording in the comment.

src/tabpfn_extensions/rf_pfn/sklearn_based_random_forest_tabpfn.py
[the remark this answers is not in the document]
safaricd: I agree, and this was a deliberate choice so that we don't need to implement two separate methods for the same purpose - `set_extension_for_function` and `set_extension_for_class`. My subjective opinion is that just `set_extension` is much cleaner 😊

src/tabpfn_extensions/unsupervised/unsupervised.py
noahho: It's not about the name but about the subgrouping of functions. We have subfunctions which now fall all under unsupervised so we loose the information if a user is e.g. doing data generation or getting embeddings. This would be valuable to tarck and you'd need to move th tracking calls further down the call chain I believe

src/tabpfn_extensions/__init__.py
noahho: That would be perfect

safaricd: Changed the wording to make it more transparent

src/tabpfn_extensions/unsupervised/unsupervised.py
safaricd: Gotcha! Here's what I propose (and have already added):
- `unsupervised:embeddings`
- `unsupervised:impute`
- `unsupervised:outliers`
- `unsupervised:synthetic`

This allows us within PostHog to gather data on both general unsupervised usage + usage across specific functions.

src/tabpfn_extensions/hpo/tuned_tabpfn.py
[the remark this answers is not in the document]
safaricd: Changed that.

src/tabpfn_extensions/unsupervised/unsupervised.py
noahho: that sounds perfect even though I don't see it added in the code yet?

noahho (changes requested):
Hey can you please push the updated code? some things you say are resolved but I don't actually see in the code

src/tabpfn_extensions/unsupervised/unsupervised.py
safaricd: Yes, it is part of the node - e.g. https://github.com/PriorLabs/tabpfn-extensions/pull/162/files#diff-069f63ce8186b490ce98b31e67c825934142087f5c73c70265283bb61f66631fR536

src/tabpfn_extensions/rf_pfn/sklearn_based_random_forest_tabpfn.py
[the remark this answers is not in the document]
noahho: Agreed about "set_extension_for_function and set_extension_for_class"
At the same time that's not the point of this comment. We should add set_extension to RandomForestTabPFNClassifier, and RandomForestTabPFNRegressor instead of the functions as done here?

PR #166, opened 2025-09-17
Run the tests on push to main, and also when requested.

oscarkey:
At the moment we only run the tests in PRs, but this can mean that the tests are broken after merge to main due to interactions with concurrent PRs.

Also add a "workflow_dispatch" trigger so we can run the tests manually.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -3,6 +3,9 @@ on:
   pull_request:
     branches:
       - main
+  push:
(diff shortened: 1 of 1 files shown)


oscarkey:
@SilenceX12138 yes this could be a good idea. Imo we should start with this check, and enable the additional protection (and thus additional overhead) only if we see a lot of problems.

safaricd (approved):
Great initiative!

PR #168, opened 2025-09-18
Interpretability: Partial Dependence + ICE Plots 

klemens-floege:
Enables creation of Partial Dependence Plot using basic Sklearn API: 

pdp


noahho (approved):
Cool stuff!

PR #190, opened 2025-10-27
[WIP] Survival robert

noahho:
## 🚀 Add Survival Analysis Extension (`TabPFNSurvivalAnalysis`)

This PR introduces a new extension for survival analysis by integrating TabPFN with the `scikit-survival` library. It adds a new `TabPFNSurvivalAnalysis` class that ensembles a `TabPFNClassifier` and a `TabPFNRegressor` to predict risk scores, optimizing for survival analysis metrics like the concordance index.

### Description

The new `TabPFNSurvivalAnalysis` class inherits from `sksurv.base.SurvivalAnalysisMixin` to provide a `scikit-survival`\-compatible API.

The core idea is to blend two different predictions:

1.  **Event Prediction:** A `TabPFNClassifier` (`cls_model`) is trained on the event indicator (e.g., `True` for event occurred, `False` for censored).
    
2.  **Risk/Time Prediction:** A `TabPFNRegressor` (`reg_model`) is trained _only_ on samples where the event occurred (`y_event is True`). The target for this regressor is not the raw time, but a **ranked risk score**. This score is derived by ranking the event times (from longest to shortest) and normalizing them between 0.0 and 1.0.
    

The final `predict` method returns a combined risk score by multiplying the classifier's probability of an event (`predict_proba`\[..., 1\]) by the regressor's predicted risk score.

### Key Changes

-   **New Module (`src/tabpfn_extensions/survival/`)**:
    
    -   Adds `sksurv_based_tabpfn_survival_analysis.py` containing the new `TabPFNSurvivalAnalysis` class.
        
    -   Adds `__init__.py` to export the class.
        
-   **Dependencies (`pyproject.toml`)**:
    
    -   Adds a new optional dependency group `[survival_pfn]`.
        
    -   This group requires `scikit-survival>=0.25.0` (for Python >= 3.10).

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -7,6 +7,10 @@ on:
     branches: [main]
   workflow_dispatch:
 
+concurrency:
+  group: pr-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   check_python_linting:
     name: Ruff Linting & Formatting
@@ -16,11 +20,11 @@ jobs:
       - uses: astral-sh/ruff-action@57714a7c8a2e59f32539362ba31877a1957dded1 # v3.5.1
         with:
           src: "./src ./tests"
-          version: 0.8.6
+          version-file: pyproject.toml
       - uses: astral-sh/ruff-action@57714a7c8a2e59f32539362ba31877a1957dded1 # v3.5.1
         with:
           src: "./src ./tests"
-          version: 0.8.6
+          version-file: pyproject.toml
           args: "format --check"
 
   test_compatibility:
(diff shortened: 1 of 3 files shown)


src/tabpfn_extensions/survival/sksurv_based_tabpfn_survival_analysis.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def __init__(
+        self,
+        *,
+        cls_model_path="auto",
+        reg_model_path="auto",
+        ignore_pretraining_limits=False,
+        random_state=None,
+    ):
+        self.cls_model = TabPFNClassifier(
```
jediroberthatch: Would it be better to pass in 2 tabpfn model instances, and have the caller initialize the TabPFNClassifier and TabPFNRegressor? More like how rf_pfn does it?

PR #192, opened 2025-10-31
Fix the ruff configuration.

oscarkey:
So that the pre-commit uses the same rules as ruff run manually and in the CI.

Also disable unsafe fixes.


priorphil (approved):
Thanks! Not for this PR, but I guess at some point we should also upgrade the ruff version here :)

PR #193, opened 2025-10-31
Remove dead code + update search space.

priorphil:


LeoGrin (approved):
Not familiar with the original code but it LGTM thanks!

I wrote on slack that adding `MIN_UNIQUE_FOR_NUMERICAL_FEATURES=[1, 5, 10, 30]` to the search space would be a bit too experimental but thinking about it more I'm quite hopeful it would help to add it.

bejaeger (approved):
I think given that we're seeing the different behavior when fixing (and thus changing) the handling of numerical and categorical indices, I think I agree that we should probably add what Leo suggested.

Otherwise LGTM!

PR #204, opened 2025-11-11
Add new model paths and version selector to AutoTabPFN and tuning.

priorphil:


src/tabpfn_extensions/hpo/search_space.py
```diff
@@ -119,36 +134,43 @@ def get_param_grid_hyperopt(task_type: str) -> dict:
...
+    model_paths = [str(model_dir / ckpt_name) for ckpt_name in model_source.filenames]
     search_space["model_path"] = hp.choice("model_path", model_paths)
 
+    search_space["inference_config/REGRESSION_Y_PREPROCESS_TRANSFORMS"] = hp.choice(
+        "REGRESSION_Y_PREPROCESS_TRANSFORMS",
+        [
+            (None,),
+            (None, "safepower"),
+            ("safepower",),
+            # ("quantile_uni",),
```
klemens-floege: is this meant to commented?

priorphil: Yes, we have a bunch of commented options that can be enabled if you have infinite GPU and want the best performance (explained in README).

Will move it under `if regression` though.

PR #207, opened 2025-11-12
Add multi-output classifier support and refresh examples

noahho:
## Summary
- extend the multi-output helper to cover both regression and classification by sharing TabPFN initialisation logic
- refresh the multi-output example to demonstrate both tasks without emphasising imputation in the naming
- expand the regression test and add a classification variant to ensure missing value handling works for either backend

## Testing
- FAST_TEST_MODE=1 pytest tests/test_multioutput.py

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69139cc7c0ac8333b836e3b79dffd925)


examples/multioutput/multioutput_prediction.py
```diff
@@ -0,0 +1,74 @@
...
+X_clf, y_clf = make_multilabel_classification(
+    n_samples=150,
+    n_features=6,
+    n_classes=3,
+    n_labels=2,
+    allow_unlabeled=False,
+    random_state=1,
+)
+
+X_clf = X_clf.astype(np.float32)
```
bejaeger, line 61: why convert here but not for regression above?

src/tabpfn_extensions/multioutput.py
```diff
@@ -0,0 +1,75 @@
+"""Utilities for multi-output learning with TabPFN."""
```
bejaeger:
```suggestion
"""Wrapper for multi-output learning with TabPFN."""
```

src/tabpfn_extensions/multioutput.py
```diff
@@ -0,0 +1,75 @@
...
+            estimator = self._tabpfn_estimator_cls(**tabpfn_params)
+
+        super().__init__(estimator=estimator, n_jobs=n_jobs)
+
+    def get_params(
+        self, deep: bool = True
+    ) -> dict[str, Any]:  # pragma: no cover - delegating to sklearn
+        """Return parameters for this estimator with TabPFN kwargs included."""
+        params = super().get_params(deep=deep)
+        if getattr(self, "_estimator_is_default", False):
```
bejaeger:
```suggestion
        if self._estimator_is_default:
```
We always set the variable so would be a bit more readable.

src/tabpfn_extensions/multioutput.py
```diff
@@ -0,0 +1,75 @@
...
+class _TabPFNMultiOutputMixin:
+    """Shared initialisation logic for TabPFN multi-output wrappers."""
+
+    _tabpfn_estimator_cls: type[_EstimatorT]
+
+    def __init__(
+        self,
+        estimator: _EstimatorT | None = None,
+        *,
+        n_jobs: int | None = None,
```
bejaeger:
```suggestion
        n_preprocessing_jobs: int | None = None,
```
We are deprecating `n_jobs` right now. Please change to `n_preprocessing_jobs` across the board.

bejaeger (approved):
Nice! Just have a few nits. Otherwise LGTM!

noahho:
Closing — decided we don't want a dedicated multi-output wrapper module. The `_TabPFNMultiOutputMixin` machinery (~75 LOC + `get_params`/`set_params` plumbing) is doing a lot of work for what's a 2-line user pattern:

```python
from sklearn.multioutput import MultiOutputRegressor
model = MultiOutputRegressor(TabPFNRegressor())
```

The actual friction — users not knowing this pattern exists — is solved better by an example file than a wrapper class. Replaced by an example at https://github.com/PriorLabs/tabpfn-extensions/pull/297.

PR #217, opened 2025-11-24
Tidy up dependency management and CI.

oscarkey:
This will allow us to bump the minimum version newer than that pushed to PyPI, to fix compatibility issues. I'm unsure if this is a good idea, what do you think?

Also, tidy up a bit:
- Switch to uv min/max dependency resolution, rather than our script
- Bump shap minimum version to get min dependency resolution to work (it's still 15 months old)
- Don't duplicate dependencies in "all" optional dependency, and reference other optional dependency sets instead
- Use "ci" dependency group, rather than puting depedencies in workflow file
- Merge windows and linux jobs
- Check in uv.lock


oscarkey:
covered by https://github.com/PriorLabs/tabpfn-extensions/pull/265

Issue #216, opened 2025-11-24
Bug with string type conversion in `ManyClassClassifier` extension


MaxSchambach:
### Describe the bug

Due to weird behaviour of scikit-learn's `check_array`, feature arrays with string features cause an error in `validate_data` in the many class extension.

I had appearingly randomly failing cases for tables that involving string data types.
I was able to track it down to string values, but the error only appears when no other non-standard dtypes are present and if they are kepy as native string and not, e.g., category dtypes. For example. in combination with bools or categories, the error disappears, see code and output below.

The root of this in the missing/incorrectly set dtype in `validate_data` respectively `check_array`.
As TabPFN can natively handle strings, categorical, bools and others, the default `dtype='numeric'` is incorrect in my opinion and leads to `could not convert string to float: 'cat1'` errors in `check_array`.

**The fix is easy:** just add `dtype=None` to the `validate_data` call [here](https://github.com/PriorLabs/tabpfn-extensions/blob/main/src/tabpfn_extensions/many_class/many_class_classifier.py#L191) and [here](https://github.com/PriorLabs/tabpfn-extensions/blob/main/src/tabpfn_extensions/many_class/many_class_classifier.py#L258). (Assuming that feature encoding is handled by the base estimator, TabPFN in this case, and not expected to be performed by the user beforehand).

I tried with both `scikit-learn` 1.6.1 and 1.7.2 (so not using the many class extension patch of `validate_data`.

Code to reproduce:

```python
from tabpfn.classifier import TabPFNClassifier
from tabpfn_extensions.many_class import ManyClassClassifier


def get_X_y(
        use_category:  bool = False,
        use_bool: bool = False,
        num_rows: int = 100,
        num_target_classes: int = 3,
):
    """Create a mock dataframe that contains floats, strings/cats and potentially bool values."""
    import numpy as np
    import pandas as pd
    import random

    assert num_rows > 5

    def _random_from_cats(cats, num_rows):
        """Sample random values from a list of categories but guarantee that all categories are present in the result."""
        return cats + random.choices(cats, k=num_rows - len(cats))

    data = {}

    # Create pandas series with low cardinality categorical values
    _categories = ["cat1", "cat2", "cat3"]
    _series = pd.Series(_random_from_cats(_categories, num_rows), name="cat_low_cardinality_str")
    data["cat_low_cardinality_str"] = _series
    if use_category:
        _series = pd.Series(
[...]

PR #218, opened 2025-11-25
Fixed `validate_data` for `ManyClassClassifier` when using string inputs.

adrian-prior:
This PR fixes the issues reported in #216 by correcting the `dtype` assertion in `validate_data` to `None`. This allows string columns to be passed directly to the underlying estimator, which works with TabPFN.

I also created a test case based on the example code provided by the user to make sure that we pass through string columns in the future.

### DIFF
--- a/tests/test_many_class_classifier.py
+++ b/tests/test_many_class_classifier.py
@@ -329,7 +329,7 @@ class TestManyClassClassifier(BaseClassifierTests):  # Inherit from BaseClassifi
         pass
 
     def test_with_pandas_and_mixed_datatypes(
-        self, estimator, pandas_classification_dataset_many_classes_mixed_types
+        self, pandas_classification_dataset_many_classes_mixed_types
     ):
         # Use a fast dummy classifier instead of TabPFN for speed; cannot use standard estimator due to mixed types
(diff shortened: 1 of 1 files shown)


tests/test_many_class_classifier.py
```diff
@@ -336,3 +410,68 @@ def test_with_pandas(self, estimator, pandas_classification_data):
     )
     def test_with_missing_values(self, estimator, dataset_generator):
         pass
+
```
noahho: I think removing "pass" from def test_with_pandas(self, estimator, pandas_classification_data) above might do jst what you wanted. This is a shared test that was disabled for ManyClass as it wasn't compatible before.

noahho: Can you check that that - would reduce complexity if can be shared

noahho (changes requested):
Please check the comment

tests/test_many_class_classifier.py
adrian-prior: Thanks for pointing this out! I couldn't reuse the same fixture, but I could reuse the data-generation code we already use, which reduced the diff significantly. Is it better like this?

PR #235, opened 2026-01-17
standardize the use of random state across random forest

### LINKED ISSUE #211: Allow random state control in RF PFN.
### Describe the workflow you want to enable

The random state handling is very suboptimal in the current API. When instantiating random forest, we don't overwrite the tree_seed (https://github.com/PriorLabs/tabpfn-extensions/blob/main/src/tabpfn_extensions/rf_pfn/sklearn_based_decision_tree_tabpfn.py#L150)  If the tree seed is 0, it's getting randomized: https://github.com/PriorLabs/tabpfn-extensions/blob/main/src/tabpfn_extensions/rf_pfn/sklearn_based_decision_tree_tabpfn.py#L313-L314.

Another confusing thing is that you can pass a random_state but if we use tabpfn that is ignored https://github.com/PriorLabs/tabpfn-extensions/blob/main/src/tabpfn_extensions/rf_pfn/sklearn_based_decision_tree_tabpfn.py#L219-L220

### Describe your proposed solution

At the very least, allow setting the tree seed. Ideally pass on the random state if passed instead. Haven't thought deeply about this yet though.

### Describe alternatives you've considered, if relevant

_No response_

### Additional context

_No response_

### Impact

None

poonai:
This fixes #211.

The current implementation ignores the user-provided random seed and instead initializes a unique random state for each leaf node. However, since leaf nodes already operate on randomly sampled data, maintaining separate random states per leaf provides no meaningful benefit.

This change uses the user-specified random state uniformly across all leaf nodes, removing unnecessary complexity and simplifying the code.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -61,13 +61,13 @@ jobs:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: ${{ matrix.python-version }}
           architecture: x64
 
       - name: Install uv
-        uses: astral-sh/setup-uv@ed21f2f24f8dd64503750218de024bcf64c7250a # v7.1.5
+        uses: astral-sh/setup-uv@803947b9bd8e9f986429fa0c5a41c367cd732b41 # v7.2.1
         with:
           enable-cache: true
 


tests/test_rf_pfn.py
```diff
@@ -77,3 +111,59 @@ def test_with_text_features(self, estimator, dataset_generator):
...
+    X, y, estimator_same_seed_1, estimator_same_seed_2, estimator_diff_seed
+):
+    # Get decision paths for both instances with same random_state
+    decision_path_1 = estimator_same_seed_1.estimators_[0].decision_path(X)
+    decision_path_2 = estimator_same_seed_2.estimators_[0].decision_path(X)
+
+    # Assert same random_state produces same decision paths
+    assert (
+        decision_path_1.toarray() == decision_path_2.toarray()
+    ).all(), "Same random_state should produce identical decision paths"
```
adrian-prior: Could you also check that the `random_states` in all the leave nodes match (for the same random seed)?

adrian-prior: Also, why not check the decision paths across all estimators?

adrian-prior (commented):
Thanks a lot for taking on this issue! There is a minor typo in one of the docstrings (as flagged by Gemini) and I left a minor comment.

I think you made/didn't change one subtle design decision here, that I am unsure about: Every tree in the random forest now uses the same random seed (if fixed). While we do use different random seeds for bootstrapping each of the subsets, I do believe that we could gain additional diversity by (deterministically) varying the seed from tree to tree. Additionally, we fix the same random seed in each leaf of a decision tree, but I am less worried about this. Wdyt?

tests/test_rf_pfn.py
```diff
@@ -77,3 +117,104 @@ def test_with_text_features(self, estimator, dataset_generator):
...
+    rf_clf : RandomForestTabPFNClassifier or RandomForestTabPFNRegressor
+        A fitted random forest estimator with multiple estimators.
+    X : array-like
+        The input samples to compute decision paths for.
+    """
+    # Collect all decision paths
+    decision_paths = [
+        rf_clf.estimators_[i].decision_path(X).toarray()
+        for i in range(len(rf_clf.estimators_))
+    ]
```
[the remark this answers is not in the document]
poonai: not required for test case

tests/test_rf_pfn.py
poonai: 2nd comment: addressed your suggestion. 

1st comment: The current changes will set different random state across leaf node. After receiving your thoughts, I checked the scikit source code. They set different random state for each tree to improve the diversity of split generation. Therefore, I kept the old logic as it is. Except, I organized it and made the `tree_seed` to work. Additionally, I added the missing test case. 

Thanks for spending time to review the PR.

tests/test_rf_pfn.py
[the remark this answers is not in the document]
adrian-prior: Doesn't this just make the code more efficient?

tests/test_rf_pfn.py
```diff
@@ -77,3 +117,104 @@ def test_with_text_features(self, estimator, dataset_generator):
...
+            assert not equal_decision_path(decision_paths[i], decision_paths[j])
+
+
+def equal_decision_path(path_a, path_b):
+    # Helper function to compare two decision paths for equality.
+    if path_a.shape != path_b.shape:
+        return False
+    path_a_dense = path_a.toarray() if hasattr(path_a, "toarray") else path_a
+    path_b_dense = path_b.toarray() if hasattr(path_b, "toarray") else path_b
+    return np.array_equal(path_a_dense, path_b_dense)
```
adrian-prior: If you're already casting the decision paths to arrays above, this helper seems a bit unnecessary.

tests/test_rf_pfn.py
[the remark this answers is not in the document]
poonai: Hi, 

Initially it was not intuitively understandable and I also was lazy to add optimization for test case. It was my mistake. 

I addressed it.

PR #238, opened 2026-01-23
Add license checks

simo-prior:


.github/workflows/pull_request.yml
```diff
@@ -78,12 +78,24 @@ jobs:
...
           fi
         shell: bash
 
+      - name: Check for forbidden licenses
+        if: runner.os == 'MacOS' && matrix.python-version == '3.9'
+        run: |
+          licensecheck \
+            --requirements-paths pyproject.toml \
+            --only-licenses APACHE MIT BSD ISC PYTHON UNLICENSE UNKNOWN \
+            --ignore-packages certifi "tabpfn*" \
```
bejaeger, line 95: Not sure if we want to be consistent but in `tabpfn` we defined the ignored packages in pyproject with some comments:
https://github.com/PriorLabs/TabPFN/blob/main/pyproject.toml#L412

pyproject.toml
```diff
@@ -103,6 +103,7 @@ dev = [
     "pytest-mock>=3.15.1",
     "build>=1.3.0",
     "twine>=6.2.0",
+    "licensecheck==2025.1.0",
```
bejaeger:
```suggestion
    "licensecheck>=2025.1.0",
```
to be consistent with `tabpfn`?

bejaeger (approved):
Thanks!

.github/workflows/pull_request.yml
simo-prior: @bejaeger Ah right, I went for `tabpfn-client`'s approach https://github.com/PriorLabs/tabpfn-client/blob/main/.github/workflows/pull_request.yml#L107.

I think inlining makes things clearer, especially for people who get this failure for the first time and need to debug.

PR #257, opened 2026-03-27
[WIP] perf: parallelize ManyClassClassifier with multi-GPU dispatch and X-preprocessing cache

julienlafrance:
## Problem

`ManyClassClassifier.predict_proba()` runs all ECOC sub-problems sequentially on a single GPU:

```python
Y_pred_probas_list = [
    _fit_and_predict_proba(self.estimator, self.X_train,
                           self.Y_train_per_estimator[i], X)
    for i in tqdm.tqdm(range(self.code_book_.shape[0]))
]
```

Each call to `_fit_and_predict_proba` clones the estimator, runs `fit()` (which repeats the full X preprocessing), then `predict_proba()`. For a 82-class problem, this means ~80 sequential sub-problems × ~880ms each ≈ **70 seconds on a single GPU**.

## Root cause

Profiling shows the per-sub-problem cost breaks down as:

| Component | Time | Notes |
|---|---|---|
| `fit_transform_ensemble_members` | ~500ms | 8 ensemble members × sklearn ColumnTransformer on X |
| `create_inference_engine` | ~50ms | GPU tensor setup |
| GPU forward pass | ~330ms | Actual transformer inference |

**The 500ms preprocessing of X is identical across all sub-problems** because `preprocessor.fit_transform(X_train, feature_schema)` does not take `y` as input (`preprocessing/transform.py:67`). Labels are handled separately via `class_permutation[y]` (line 99). This means 79 out of 80 preprocessing runs are pure waste.

## Solution

Two independent optimizations, each backward-compatible:

**1. Multi-GPU parallel dispatch (`n_jobs` parameter)**

Add a `n_jobs` parameter to `ManyClassClassifier`. Sub-problems are dispatched to persistent worker processes (one per GPU) via `torch.multiprocessing`. Each worker loads TabPFN once at startup and processes its batch of sub-problems sequentially.

- Speedup: **÷n_gpus** (linear scaling, sub-problems are fully independent)
- No accuracy change (verified on 18 datasets across 4 CV folds, zero regressions)

**2. X-preprocessing cache (y-swap)**

Since X preprocessing is independent of y, the first sub-problem runs a full `fit()`, then subsequent sub-problems reuse the cached `executor_.ensemble_members` and only replace `y_train`:

```python
# First sub-problem: full fit (880ms)
cached = clone(clf)
cached.fit(X_train, y_first)

# Subsequent sub-problems: swap y only (~380ms)
for em in cached.executor_.ensemble_members:
    perm = em.config.class_permutation
    y_new = perm[y_bin] if perm is not None else y_bin
    em.y_train = torch.tensor(y_new, dtype=torch.long, device=em.y_train.device)
proba = cached.predict_proba(X_test)
```

- Speedup: **×1.9 per sub-problem** (880ms → 469ms)
- No accuracy change (X preprocessing and GPU forward pass are identical; only the label vector changes)

## Combined results

Benchmarked on 18 classification tasks across 4 cross-validation folds (13-82 classes, numerical features, 2× RTX 3090):

| Configuration | Time (82 classes) | Speedup |
|---|---|---|
| Sequential (baseline) | ~58s | 1× |
| Multi-GPU pool + y-swap (2 GPUs) | **8.3s** | **7.0×** |

Full benchmark (18 datasets, 4 folds, 2× RTX 3090):

| Classes | Sequential time | Pool + y-swap | Speedup | Accuracy delta |
|---|---|---|---|---|
| 82 | 57-59s | 8.3-8.4s | 6.8-7.0× | 0.0% |
| 48-55 | 29-56s | 4.8-11.4s | 4.9-6.6× | 0.0% to +0.2% |
| 32-44 | 23-171s | 3.5-41.8s | 4.1-6.5× | 0.0% to +1.7% |
| 13-26 | 9-21s | 1.7-3.9s | 3.5-5.7× | 0.0% |

Aggregate over all 18 datasets:

| Metric | Sequential | Pool + y-swap |
|---|---|---|
| Total time | 625s | 120s |
| Mean time per dataset | 34.7s | 6.7s |
| Mean speedup | — | **5.6×** |
| Mean accuracy | 94.1% | 94.3% |
| Max accuracy regression | — | **none (0 of 18)** |

**Accuracy is identical or marginally better on all 18 datasets. Zero regressions observed.**

## Why this matters

TabPFN's strength on many-class problems (>10 classes) is gated by the ManyClassClassifier wrapper performance. Users with multi-GPU setups or large-scale classification tasks currently hit a wall where inference time scales linearly with class count. These two optimizations are orthogonal, require no changes to the TabPFN model itself, and maintain full backward compatibility.

## Changes

- **`many_class_classifier.py`**: Add `n_jobs` and `cache_preprocessing` parameters. When `n_jobs > 1` and `start_pool()` has been called, dispatch sub-problems to persistent GPU workers instead of sequential loop.
- **`_parallel.py`** (new): Persistent worker process with y-swap logic.
- **`test_parallel_many_class.py`** (new): 6 tests validating parallel dispatch, probability normalization, fallback behavior, cache toggle, no-mapping path, and diagnostics.

## Backward compatibility

- Default behavior unchanged (`n_jobs=1`, `cache_preprocessing=True`)
- `cache_preprocessing=True` is safe by default since X preprocessing is provably independent of y
- No changes to TabPFN core model, only to the `tabpfn-extensions` wrapper

## Test plan

- [x] 6 unit tests passing (DecisionTreeClassifier as base estimator, no GPU required)
- [x] Benchmark on 18 industrial datasets with TabPFN V2 on 2× RTX 3090
- [ ] CI validation on PriorLabs infrastructure

### DIFF
--- a/src/tabpfn_extensions/many_class/_parallel.py
+++ b/src/tabpfn_extensions/many_class/_parallel.py
@@ -1,0 +1,173 @@
+"""Multi-GPU parallel dispatch for ManyClassClassifier.
+
+Persistent workers load TabPFN once per GPU and process batches of ECOC
+sub-problems. Includes y-swap optimization: first sub-problem runs full fit()
+(caching X preprocessing), subsequent sub-problems only replace y_train in
+cached ensemble members, skipping ~500ms of redundant preprocessing per row.
+
+This is safe because preprocessor.fit_transform(X_train, feature_schema) does
+not take y as input (see preprocessing/transform.py). Labels are handled
+separately via config.class_permutation[y].
+"""
+from __future__ import annotations
+
+import os
+from typing import Any
+
+import numpy as np
+import torch.multiprocessing as mp
+from sklearn.base import clone
+
+from ._utils import (
+    EPS_WEIGHT,
+    RowRunResult,
+    align_probabilities,
+    apply_categorical_features,
+    as_numpy,
+)
+
+
+def _worker(gpu_id: int, task_queue: mp.Queue, result_queue: mp.Queue) -> None:
+    """Persistent worker: load TabPFN once, process batches with y-swap."""
+    import torch
+
+    os.environ["CUDA_VISIBLE_DEVICES"] = str(gpu_id)
+
+    from tabpfn import TabPFNClassifier
+    from tabpfn.constants import ModelVersion
+
+    template = TabPFNClassifier.create_default_for_version(ModelVersion.V2)
+    template.ignore_pretraining_limits = True
+
+    result_queue.put({"status": "ready", "gpu_id": gpu_id})
+
+    while True:
+        msg = task_queue.get()
+        if msg is None or msg.get("cmd") == "stop":
+            result_queue.put({"status": "stopped", "gpu_id": gpu_id})
+            break
+
+        X_train = msg["X_train"]
+        X_test = msg["X_test"]
+        rows = msg["rows"]  # list of (row_idx, y_codes, mask_or_None)
+        alphabet_size = msg["alphabet_size"]
+        categorical_features = msg.get("categorical_features")
+        fit_params = msg.get("fit_params") or {}
+        cache_preprocessing = msg.get("cache_preprocessing", True)
+
+        results: dict[int, RowRunResult] = {}
+        cached = None
+
+        for idx, (row_idx, y_codes, mask) in enumerate(rows):
+            X_row = X_train[mask] if mask is not None else X_train
+            y_row = y_codes[mask] if mask is not None else y_codes
+
+            if len(y_row) == 0:
+                n_test = as_numpy(X_test).shape[0]
+                results[row_idx] = RowRunResult(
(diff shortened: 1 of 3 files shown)


alanprior:
Hello @julienlafrance and thank you so much for this fix! Nice finding!

We are currently working on some related improvements, so I'm a bit hesitant regarding prioritization of handling this PR very soon, but most definitely we should make multi-class inference much faster and it will happen in the near future! :)

cc @noahho for visibility.

adrian-prior:
@julienlafrance, with the latest version of TabPFN (TabPFN-3), we have introduced GPU-based preprocessing and a much higher class limit of 160 classes. This should alleviate the problems described in this PR, because (a) the higher class limit means that we need to create fewer (if any subproblems) and (b) the preprocessing per estimator now runs a lot quicker.

Therefore, I will close this PR for now. If the issue still persists, please open a new issue :) Many thanks for the efforts on this PR!

PR #265, opened 2026-04-14
Switch to uv min/max dependency resolution.

oscarkey:
Rather than our custom script. This matches what we do in TabPFN.

- Add a CI dependency group, containing the minimum set of dependencies required on the CI
- Define the `all` extra set using the other extras, rather than repeating the dependencies
- Bump some minimum versions to get `lowest-direct` resolution to work: tabpfn 6.0.5 -> 7.0, llvmlite -> 2 years old, pandas -> 2 years old
- Add a `numba` direct dependency to the `interpretability` extra to fix lowest-direct.

Once this PR is ready I will update the branch protection rules to match the new job names.

Part of PRI-261


.github/workflows/pull_request.yml
```diff
@@ -105,12 +92,13 @@ jobs:
         env:
           HF_TOKEN: ${{ secrets.HF_TOKEN }}
           TABPFN_TOKEN: ${{ secrets.TABPFN_TOKEN }}
-        run: python scripts/download_all_models.py
+        run: uv run --no-sync python scripts/download_all_models.py
 
       - name: Check for forbidden licenses
         if: runner.os == 'MacOS' && matrix.python-version == '3.9'
```
adrian-prior: Hey,
I think you might need to bump the Python version here to 3.11. We don't include 3.9+macOS in the matrix anymore :)

adrian-prior (approved):
Thank you so much! Will this also prevent unnecessary upper bounds from being reintroduced?

Except the comment, everything looks good to me!

.github/workflows/pull_request.yml
oscarkey: Do you know why this check is here? Is it just to only run the license check once?
To that end I changed it to the following to try and make it more future proof:
`if: runner.os == 'MacOS' && matrix.dependency-set == 'lowest-direct'`

oscarkey:
> Will this also prevent unnecessary upper bounds from being reintroduced?

sadly not :( but it does avoid other bugs we've had with the custom script

.github/workflows/pull_request.yml
adrian-prior: That also would have been my guess, but I sadly don't know it for sure.

Thanks for fixing it!

PR #272, opened 2026-04-21
[PRI-269] AutoTabPFN diagnostics + per-checkpoint AutoGluon limits

adrian-prior:
## Summary

Motivated by investigating PRI-269 (`AutoTabPFN` silently fails with `RuntimeError: No models were trained successfully`). Three related fixes:

1. **Disable AutoGluon's v2-era static pretraining-limit guardrails for TabPFNv2 sub-models.** AutoGluon's `TabPFNV2Model._get_default_auxiliary_params` hardcodes `max_rows=10000 / max_features=500 / max_classes=10`, tuned for v2. These are wrong for v2.5 (50K / 2000) and v2.6 (100K / 2000). Previously, without `ignore_pretraining_limits=True`, AutoTabPFN would reject any dataset >10K rows *regardless* of which checkpoint was loaded. `prepare_tabpfnv2_config` now sets `ag_args_fit = {"max_rows": None, "max_features": None, "max_classes": None}` — disabling the AG-side asserts entirely and delegating gating to TabPFN's own validation (which knows the actual loaded checkpoint's capability via `inference_config_`). `ignore_pretraining_limits` still controls whether TabPFN enforces or bypasses its own check.

2. **Bump default fit verbosity from the inherited `1` to `2`.** At verbosity=1 AutoGluon suppresses the per-sub-model `Time limit exceeded... Skipping` lines — the only useful signal when the zero-models-trained failure mode fires. Added a comment explaining why the default should not be dropped back to 1.

3. **Fix the unsupported-version error in `hpo/search_space.py`.** The raise had an f-string bug — the second string literal lost its `f` prefix, so `{model_version}` was printed as the literal text instead of being interpolated. Also switched from `ValueError` to `NotImplementedError` — more precise, since no search space has been *defined* for that version rather than the argument being invalid.

Docstrings on the three AutoTabPFN classes updated to describe the new per-checkpoint behavior.

## Test plan

Verified on H100 with 40K × 120 synthetic binary classification at `max_time=60`, no `ignore_pretraining_limits=True`:

- [x] **V2.5 checkpoint (50K limit)**: before → immediate `AssertionError: ag.max_rows=10000 for model 'TabPFNv2_X'`, zero models even attempted, exits in ~10s. After → fit proceeds into AutoGluon's training loop; 60s budget is insufficient on 40K rows so sub-models skip with "Time limit exceeded... Skipping", which is the now-surfaced interpretable diagnostic.
- [x] **V2 checkpoint (10K limit)**: TabPFN's own `TabPFNValidationError` fires cleanly, with a message referencing the *actual* v2 10K limit.
- [x] f-string in `hpo/search_space.py` interpolates `model_version` correctly (and raises `NotImplementedError`).
- [x] CI (full test suite)

## Notes

- `search_space_func` is back to its original shape — no ckpt reads or per-config lookups, so no runtime overhead. AG-side asserts are neutralized with 3 `None` values in `ag_args_fit`; nothing else changes.
- Works regardless of [PriorLabs/TabPFN-private#573](https://github.com/PriorLabs/TabPFN-private/pull/573)'s merge status — we no longer rely on `InferenceConfig.get_default` at all.

### DIFF
--- a/src/tabpfn_extensions/hpo/search_space.py
+++ b/src/tabpfn_extensions/hpo/search_space.py
@@ -163,9 +163,9 @@ def get_param_grid_hyperopt(
         )
 
     else:
-        raise ValueError(
-            f"Unknown combination of task type {task_type} and "
-            "model version {model_version}!"
+        raise NotImplementedError(
+            f"No hpo search space is defined for task type {task_type} and "
+            f"model version {model_version}."
         )
 
     # Make sure models are downloaded.
--- a/src/tabpfn_extensions/post_hoc_ensembles/sklearn_interface.py
+++ b/src/tabpfn_extensions/post_hoc_ensembles/sklearn_interface.py
@@ -80,9 +80,14 @@ class AutoTabPFNBase(BaseEstimator):
         The number of internal transformers to ensemble within each individual TabPFN model.
         Higher values can improve performance but increase resource usage.
     ignore_pretraining_limits : bool, default=False
-        If `True`, bypasses TabPFN's built-in limits on dataset size (10000 samples)
-        and feature count (500). **Warning:** Use with caution, as performance is not
-        guaranteed and may be poor when exceeding these limits.
+        If `True`, bypasses the pretraining limits enforced on each TabPFN
+        sub-model. By default (`False`), AutoTabPFN reads the loaded checkpoint's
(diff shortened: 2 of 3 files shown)


LeoGrin (approved):
LGTM, thanks!

PR #275, opened 2026-04-24
Remove unjustified upper bounds on runtime dependency pins

adrian-prior:
## Summary

Removes the `<Y` upper bounds on `torch`, `pandas`, `scikit-learn`, `scipy`, and `tabpfn` in the runtime `dependencies` list. These caps were causing real-world breakage — most recently, `tabpfn<7` silently pulled tabpfn 6.4.1 into fresh installs of `tabpfn-extensions[all]`, which in turn made the Colab quickstart default to the v2.5 model instead of v2.6. See [PRI-280](https://linear.app/priorlabs/issue/PRI-280/colab-notebook-installs-outdated-tabpfn-version-641-pulls-v25-model) and the Slack discussion below.

The main `tabpfn` package doesn't cap these dependencies either, so this also just aligns our conventions.

### What changed

```diff
 dependencies = [
-    "torch>=2.1,<3",
-    "pandas>=1.4.0,<3",
-    "scikit-learn>=1.6.0,<1.7",
-    "scipy>=1.11.1,<2",
-    "tabpfn>=6.0.5,<8",
+    "torch>=2.1",
+    "pandas>=1.4.0",
+    "scikit-learn>=1.6.0",
+    "scipy>=1.11.1",
+    "tabpfn>=6.0.5",
     "tabpfn-common-utils[telemetry-interactive]>=0.2.0",
 ]
```

### What was kept

Upper bounds that have an accompanying comment explaining the constraint are preserved:

- `setuptools>=67.0.0,<82` (in `hpo` and `all` extras) — comment links to the pkg_resources removal transition.
- `ruff==0.8.6` (in the `dev` group) — must match `.pre-commit-config.yaml`.
- `autogluon.tabular==1.4.0` — exact pin, not a cap; being replaced with `>=1.5.0,<1.6` in #274.

### Why now

Per [Leo's approval in Slack](https://prior-labs.slack.com/archives/C08LK13JRUN/p1777061263049719?thread_ts=1776863255.568099&cid=C08LK13JRUN) in the thread about PRI-280:

> "Yes would be great to remove!!"

## Test plan
- [x] CI green (all existing tests should continue to pass — we're loosening constraints, not tightening them)


alanprior:
I think the solution is to only change "tabpfn>=6.0.5". Were the other upper limits causing any issues? New majors DO TEND TO BREAK things. 

cc @LeoGrin

alanprior:
@adrian-prior WDYT?

LeoGrin (approved):
Thanks!! Checked and the CI is testing the latest version

PR #280, opened 2026-04-28
Add support for "auroc" metric alias in classification scoring

psinger-prior:
## Summary
This change adds support for the "auroc" metric name as an alias for "roc" in the classification scoring function, improving consistency with common metric naming conventions.

## Key Changes
- Updated the ROC metric check to accept both "roc" and "auroc" as valid metric names when handling binary classification predictions
- Extended the metric validation logic to recognize "auroc" alongside "roc" and "log_loss" when determining whether to convert probability predictions to class labels
- Ensures consistent behavior whether users specify "roc" or "auroc" as their optimization metric

## Implementation Details
The changes maintain backward compatibility by treating "auroc" as an equivalent alias to "roc" throughout the scoring pipeline. This allows users to use either naming convention without affecting the underlying functionality.

This fixes other scripts running the wrong scoring metric when using "auroc", eg
https://github.com/PriorLabs/tabpfn-extensions/blob/main/src/tabpfn_extensions/hpo/tuned_tabpfn.py#L280


priorjulien (approved):
LGTM, and apologies for the late reply!

PR #282, opened 2026-05-08
Auto-infer alphabet_size in ManyClassClassifier from the checkpoint

LeoGrin:
## Summary

- `ManyClassClassifier(estimator=TabPFNClassifier(), ...).fit(X, y)` now works without requiring an explicit `alphabet_size`. It is read off the base estimator's checkpoint (`MAX_NUMBER_OF_CLASSES`) automatically.
- The previous `estimator.max_num_classes_` fallback has been dead since this repo's initial commit — TabPFN core has never set that attribute on any version. Removed.

## Resolution cascade

1. Explicit `alphabet_size=...` (unchanged).
2. `estimator.inference_config_.MAX_NUMBER_OF_CLASSES` if the base estimator was already fit.
3. Probe: clone the base estimator and fit it on 4×2 synthetic rows to populate `inference_config_`, then read the value. Cost is dominated by the checkpoint load that fit was going to pay anyway — TabPFN's `_load_checkpoint_cached` (`@lru_cache(maxsize=1)`) makes the subsequent codebook fits a cache hit. Net I/O = one ckpt read.
4. Else `ValueError` (preserves the prior contract for non-TabPFN estimators).

## Verified

Resolved `alphabet_size_` after `mcc.fit(X, y)` with no `alphabet_size` argument:

| ckpt | auto-resolved `alphabet_size_` |
|---|---|
| v2.5 default | 10 |
| v2.6 default | 10 |
| v3 default | 160 |

So for e.g. 5K-class problems the wrapper still kicks in on v3 — just with a 160-symbol alphabet (~32 base fits) instead of 10 — which is a big efficiency win over v2.x.

## Test plan

- [ ] Existing `tests/test_many_class_classifier.py` still passes.
- [ ] Smoke-test on v2.5 / v2.6 / v3 default checkpoints with no `alphabet_size` arg (covered manually during development).
- [ ] Confirm explicit `alphabet_size=N` still bypasses the probe.

---

> [!NOTE]
> **Medium Risk**
> Behavior changes in `ManyClassClassifier.fit()` by auto-inferring `alphabet_size` (and adding TabPFN-specific fallback/logging), which can alter how many base fits/codebook rows are used and thus affect performance/results across TabPFN versions.
> 
> **Overview**
> `ManyClassClassifier` no longer requires an explicit `alphabet_size` when the base estimator exposes `get_inference_config()`: it now reads `MAX_NUMBER_OF_CLASSES` from that config (with a TabPFN-namespace fallback defaulting to `10` for older versions) and only raises if no limit can be determined.
> 
> `fit()` now logs whether it will run a single base fit (*no codebook*) vs. perform output-coding over multiple base fits, and the class docstring documents the new `alphabet_size` inference behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1773458b9bcbee626ba976f2e164703d7852cc94. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/tabpfn_extensions/many_class/many_class_classifier.py
+++ b/src/tabpfn_extensions/many_class/many_class_classifier.py
@@ -240,6 +240,11 @@ class ManyClassClassifier(BaseEstimator, ClassifierMixin):
         self._row_class_mask_ = None
 
         if self.no_mapping_needed_:
+            logger.info(
+                "Base estimator supports up to %d classes; data has %d — fitting once (no codebook).",
+                self.alphabet_size_,
+                n_classes,
+            )
             estimator = self._clone_base_estimator()
             estimator.fit(X_validated, y_validated, **self.fit_params_)
             self.estimators_ = [estimator]
@@ -251,6 +256,13 @@ class ManyClassClassifier(BaseEstimator, ClassifierMixin):
             return self
 
         n_estimators = self._get_n_estimators(n_classes, self.alphabet_size_)
+        logger.info(
+            "Base estimator supports up to %d classes; data has %d — using %d-symbol output coding over %d base fits.",
+            self.alphabet_size_,
+            n_classes,
+            self.alphabet_size_,
+            n_estimators,
+        )
         codebook, stats = self._codebook_strategy.generate(
             n_classes, n_estimators, self.alphabet_size_, rng
         )
--- a/src/tabpfn_extensions/many_class/many_class_classifier.py
+++ b/src/tabpfn_extensions/many_class/many_class_classifier.py
@@ -118,6 +118,11 @@ class ManyClassClassifier(BaseEstimator, ClassifierMixin):
         X_probe = np.random.default_rng(0).standard_normal((4, 2))
         y_probe = np.array([0, 0, 1, 1])
         probe = clone(self.estimator)
+        # Strip any user-provided categorical indices — they may reference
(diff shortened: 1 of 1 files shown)


src/tabpfn_extensions/many_class/many_class_classifier.py
```diff
@@ -95,16 +105,35 @@ def _resolve_row_weighting_config(
...
+        if hasattr(probe, "categorical_features_indices"):
+            probe.categorical_features_indices = None
+        try:
+            probe.fit(X_probe, y_probe)
+        except (ValueError, TypeError):
+            # Base estimator rejected the synthetic shape/dtype — treat as
+            # "not a compatible model" and let the caller fall back to its
+            # explicit-alphabet error path.
+            return None
+        cfg = getattr(probe, "inference_config_", None)
```
adrian-prior: Couldn't this be replaced with `estimator.get_inference_config()`? I think this would be a much cleaner solution.

adrian-prior (changes requested):
Thanks for setting this up! I might be wrong, but I think there is a cleaner solution (see comment).

LeoGrin:
Yeah much better thanks! It's quite recent (Apr 24) so I added a fallback to 10 classes if we don't have this method.

adrian-prior (approved):
LGTM!

PR #283, opened 2026-05-10
[RES-1467] Drop SHAP, refresh shapiq examples around v3 KV cache

adrian-prior:
## Summary
- Remove the legacy `interpretability/shap.py` adapter and the `shap` runtime dependency. The shapiq-based path supersedes it, including for plotting (you can wrap shapiq output in `shap.Explanation` for `shap.plots.*`).
- `get_tabpfn_imputation_explainer` now (a) warns if the KV cache isn't engaged (`fit_mode="fit_with_cache"` + `executor_.keep_cache_on_device=True`), (b) defaults to `imputer="baseline"` — one forward pass per coalition, ~50× faster than `marginal` with no faithfulness loss in practice — and (c) silences shapiq's misleading "use `TabPFNExplainer` instead" warning on this path.
- Promote the KV-cache check to a shared utility in `tabpfn_extensions.utils.warn_if_no_kv_cache(model, *, context="This operation")`. Other extensions that issue many predicts per fit (HPO, feature selection, unsupervised) can drop a single line in to get the same warning, parameterized by the caller-provided `context` string.
- Rewrite `examples/interpretability/{shap,shapiq}_example.py` around California housing (regression, d=8, exact budget=2^8=256). Both examples engage the KV cache; the Rundel `get_tabpfn_explainer` path is shown commented-out with a runtime caveat.
- Refresh `interpretability/README.md` to drop the SHAP section and document the KV-cache + baseline-imputer expectations (with the Lundberg & Lee 2017 SHAP citation kept since `shap` remains in use for plotting).

## ⚠ Caveat: examples need TabPFN v3
The examples set `fit_mode="fit_with_cache"`, which is only supported by TabPFN v3. With `model_path="auto"` resolving to v2.6 today, both examples will currently error with *"fit_with_cache is not supported for TabPFN v2.6 yet."* Suggested merge order: ship this PR after the v3 release lands so `model_path="auto"` Just Works. (Locally I tested both examples against `gs://training-checkpoint/tabpfn-v3-{classifier,regressor}-20260506.ckpt`; full smoke ran end-to-end in ~67s on an RTX 6000, with all 8 expected figures rendering correctly.)

## Deferred / follow-ups
- A NaN-imputer SHAP path (re-using TabPFN's native NaN handling instead of explicit imputation) was scoped out of this PR — tracked in [PRI-290](https://linear.app/priorlabs/issue/PRI-290) which links to [PRI-287](https://linear.app/priorlabs/issue/PRI-287).
- `interpretability/experiments.py` (dead code, no callers anywhere in the repo) was *not* removed in this PR — flagged for a follow-up.

## Test plan
- [ ] Examples run cleanly under `model_path="auto"` once TabPFN v3 is released
- [ ] CI green (the dropped `shap` dep should not affect any other extra)
- [ ] Spot-check that `warn_if_no_kv_cache` (called from `get_tabpfn_imputation_explainer`) fires when constructed against a model with `fit_mode != "fit_with_cache"` or with `keep_cache_on_device=False` — verified locally on three configurations


src/tabpfn_extensions/interpretability/README.md
```diff
@@ -15,8 +15,38 @@ interactions for machine learning models. The library is a redesigned and improv
...
+
+* ``get_tabpfn_imputation_explainer`` — *imputation-based* removal (marginal / conditional /
+  baseline). The training set is fixed across coalitions, so the KV-cache fast path
+  applies. Construct your TabPFN model with ``fit_mode="fit_with_cache"`` (set BEFORE
+  ``.fit()``) and set ``model.executor_.keep_cache_on_device = True`` after fitting:
+
+  ```python
+  clf = TabPFNClassifier(fit_mode="fit_with_cache")
+  clf.fit(X_train, y_train)
+  # keep_cache_on_device is usually on by default — set explicitly as a safety net.
```
LeoGrin: don't think we need that?

examples/interpretability/shap_example.py
```diff
@@ -1,33 +1,95 @@
...
+print(f"Computing Shapley values for {n_explain} rows...")
+ivs = [explainer.explain(x=X_explain[i], budget=256) for i in range(n_explain)]
+shap_values = np.stack([iv.get_n_order_values(1) for iv in ivs])
 
-# Load example dataset
-data = load_breast_cancer()
-X, y = data.data, data.target
-feature_names = data.feature_names
-n_samples = 50
+# baseline_value is shapiq's v(empty); average across rows for a single scalar.
```
LeoGrin: not super clear what's `v(empty)`?

LeoGrin (approved):
LGTM!

A few comments from Codex I agree with
  - P1: pyproject.toml:59 still allows shapiq>=0.4.0, but the PR now defaults get_tabpfn_imputation_explainer(..., imputer="baseline") in src/tabpfn_extensions/interpretability/shapiq.py:124. With the lowest available compatible shapiq releases, baseline, index="SV", and max_order=1 fail. I reproduced this with shapiq==1.0.1; shapiq==1.1.0 works, so the lower bound should be bumped at least that far.
  - P2: pyproject.toml:58 removes shap from the interpretability extra, but examples/interpretability/shap_example.py:25 imports shap, and the README points users to shap.Explanation without a separate install note. A user installing tabpfn-extensions[interpretability] via pip will hit ModuleNotFoundError when running the advertised example. [Leo: Maybe we can just add install info to the example?]
  - P2: uv.lock:3555 was not regenerated after the dependency change; it still lists shap and numba under the interpretability extra. uv lock --check fails with “lockfile needs to be updated”.

src/tabpfn_extensions/interpretability/README.md
LeoGrin: >   # keep_cache_on_device is usually on by default — set explicitly as a safety net.

this I mean

PR #284, opened 2026-05-10
Pre-v3 cleanup: drop rf_pfn / sklearn_ensembles, deprecate AutoTabPFN+TunedTabPFN, wire KV cache

adrian-prior:
## Summary

Pre-v3 cleanup of `tabpfn-extensions`: delete dead/legacy paths, deprecate things slated for removal, and wire the KV cache (improved with TabPFN-3) into the extensions where it gives a meaningful speedup.

## Deletions
- `rf_pfn/` and `sklearn_ensembles/` packages, their examples (`examples/rf_pfn/`, `examples/large_datasets/`), and their tests.
- `rf_pfn = []` extra in `pyproject.toml`.
- The `dt_pfn` `model_type` branch from `hpo/search_space.py` + `hpo/tuned_tabpfn.py` (and the matching test arm).
- `interpretability/experiments.py` — dead code, no callers, references a `tabpfn.scripts...` path that doesn't exist.

## Deprecations
`AutoTabPFN{Classifier,Regressor}` and `TunedTabPFN{Classifier,Regressor}` keep working but emit a `DeprecationWarning` and have a banner at the top of the file. Scheduled for removal in a future release.

## KV cache

- `utils.warn_if_no_kv_cache(model, *, context)` — new helper. Detects `tabpfn-client` and recommends switching to local instead of firing a misleading "set fit_mode=..." message. Same helper as #283; whichever lands first the other rebases.
- `pval_crt.tabpfn_crt` — `use_kv_cache: bool = True` kwarg, defaults to enabled. The CRT hot loop predicts B+1 times against the same fitted model.
- `interpretability/pdp.partial_dependence_plots` — calls the warn helper for TabPFN estimators (user owns the estimator, so warn rather than mutate).
- `examples/interpretability/pdp_example.py` — constructs with `fit_mode="fit_with_cache"` and a `try/except TypeError` fallback for the client backend.

**Not wired:** `unsupervised` outlier detection (`density_()` refits per (column, permutation) — cache invalidates; verified ≈no speedup, reverted in `b50e88d`).

## Out of scope
- **SHAP / SHAPiq** — #283.
- **many_class** — could apply on v3 (~32 base fits, same X), but depends on whether the v3 cache keys on X or (X, y). Needs confirmation from the cache owner.
- **TabEBM** — uses `fit_mode="batched"` for SGLD gradients (incompatible with the cache). Also broken on `tabpfn` ≥ 7.x because `tabpfn.config` was removed (pre-existing, not a v3 regression). Tracked in [RES-1541](https://linear.app/priorlabs/issue/RES-1541/tabebm-is-broken-on-tabpfn-7x-tabpfnconfig-no-longer-exists).

## Verification

9/10 examples pass on v3 (gcp slurm, array 703369; the one failure is TabEBM per RES-1541). Measured KV-cache speedup:

| Example | No cache | With cache | Speedup |
|---|---|---|---|
| `pval_crt_demo.py` | 315s | 169s | **1.86×** |
| `pdp_example.py` | 2558s | 1291s | **1.98×** |

> All numbers CPU-only: cluster GPU drivers (570.211.01, supports CUDA 12.8) are too old for the `torch==2.11.0+cu130` wheel pulled in by `tabpfn` from `main`, so every node falls back to CPU. Real GPU speedup will be larger since the cached encoder pass is GPU-dominant.

`FAST_TEST_MODE=1 pytest tests/test_many_class_classifier.py` passes. `tests/test_hpo.py` failures pre-date this PR (reproduce on `main`).

## Cross-repo follow-ups (not in this PR)
- `docs.priorlabs.ai`: remove RF-PFN / large-datasets pages; retire or deprecate post-hoc-ensembles & HPO pages; many-class wording (default max class count moved from 10 → 160 on v3).
- Public TabPFN repo: links pointing at deleted examples / RF-PFN docs.


examples/interpretability/pdp_example.py
```diff
@@ -13,9 +15,27 @@
...
+# Initialize and train model. Enable the KV cache (improved with TabPFN-3):
+# PDP issues many predicts (grid_resolution per feature * #features) against
+# the same fitted model, so caching the encoder pass over X_train avoids
+# redoing it on every grid point. The tabpfn-client backend does not
+# currently support the KV cache, so fall back to a default constructor.
+try:
+    clf = TabPFNClassifier(fit_mode="fit_with_cache")
+except TypeError:
+    warnings.warn(
+        "PDP would benefit substantially from the KV cache, but the "
```
LeoGrin, line 32: that would also fire if you install an older tabpfn version right?

src/tabpfn_extensions/hpo/tuned_tabpfn.py
```diff
@@ -1,6 +1,13 @@
 #  Copyright (c) Prior Labs GmbH 2025.
 #  Licensed under the Apache License, Version 2.0
 
+# =============================================================================
+# DEPRECATED MODULE
+# -----------------------------------------------------------------------------
+# TunedTabPFNClassifier / TunedTabPFNRegressor are deprecated and will be
+# removed in a future release. They emit a DeprecationWarning on construction.
```
LeoGrin:
```suggestion
# removed in a future release.
```

LeoGrin (approved):
Thanks!

examples/interpretability/pdp_example.py
adrian-prior: Yeah, you are right about that. I will update the comment :)

PR #293, opened 2026-05-13
Improve feature_selection: richer return type + verbose progress

adrian-prior:
This PR improves the usability of the feature selection extensions by:
- changing the return interface
- adding verbosity so users can see what's going on (in particular, we computed some CV scores before and after, which we never logged or saved)
- Adds notes about feature selection not being that useful for TabPFN


priorjulien (approved):
LGTM.

PR #297, opened 2026-05-25
docs(examples): multi-output regression + multi-label classification

noahho:
## Summary

Adds a small example showing the standard sklearn pattern for multi-output prediction with TabPFN:

```python
MultiOutputRegressor(TabPFNRegressor())
MultiOutputClassifier(TabPFNClassifier())
```

No TabPFN-specific wrapper class — sklearn already handles the multi-output plumbing.

Covers:
- Multi-output regression (R² per target + uniform-average)
- Multi-label classification (micro-ROC-AUC), with a note about the `predict_proba` shape gotcha (`MultiOutputClassifier` returns a list of per-target arrays)

## Why an example instead of a wrapper

This replaces #207, which proposed a dedicated `TabPFNMultiOutput{Regressor,Classifier}` module (~75 lines + `get_params`/`set_params` plumbing) for what's a 2-line user pattern. Per our preference of "examples over extensions where the value-add is thin," moving this to an example makes the pattern discoverable without adding a maintained module surface.

Closes #207.

## Test plan

- [x] Syntax-check the example
- [ ] Run the example end-to-end locally (small data, n_estimators=4) — quick sanity check before merge

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -34,15 +34,13 @@ python examples/many_class/many_class_classifier_example.py
    - TabPFN Package - Full PyTorch implementation for local inference
    - TabPFN Client - Lightweight API client for cloud-based inference
 
-3. **Flexible Imports**: Use this pattern for imports to support both backends:
+3. **Imports**: Always import TabPFN classes from `tabpfn_extensions`, which
+   automatically selects the right backend (local `tabpfn` package when
+   available, falling back to `tabpfn_client`):
    ```python
-   try:
-       # Try standard TabPFN package first
-       from tabpfn import TabPFNClassifier, TabPFNRegressor
-   except ImportError:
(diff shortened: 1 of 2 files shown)


noahho:
Applied both suggestions in dade3b6 — flexible `try: from tabpfn / except ImportError: from tabpfn_client` import per the README guideline, and `r2_score(..., multioutput="raw_values")` for per-target R². Thanks!

noahho:
On the import guideline — per Noah, the actual project convention is to import directly from `tabpfn_extensions` (which handles backend selection internally). 9 of 10 existing examples already do this; the README's documented `try/except` pattern was the outlier. Reverted the example's import + updated the README to match practice (commit 531017c). Thanks for catching this.

PR #298, opened 2026-05-27
Added a scikit-learn interface to the TabPFNEmbedding object

eliott-kalfon:
## What

Replaces the bespoke `get_embeddings(X_train, y_train, X, data_source)` interface with a standard `fit` / `transform` / `fit_transform` interface.

```python
embedding = TabPFNEmbedding(n_fold=5)
train_embeds = embedding.fit_transform(X_train, y_train)  # OOF
test_embeds  = embedding.transform(X_test)                # final full-data model
```

## Why

- The old four-argument get_embeddings was awkward and didn't compose with sklearn workflows.
- A familiar fit/transform shape makes the class easier to discover and use.

### Notable behavior choices

`transform(X)` always uses the final full-data model — it does not auto-return OOF when X == X_train. Read OOF from train_embeddings_ or call fit_transform.
n_fold >= 2 with a classifier now uses `StratifiedKFold` (legacy path used plain `KFold`); deprecation message calls this out.
New shuffle: bool parameter, independent of random_state.
`model=None` auto-detects classifier/regressor from y with `n_estimators=1` and a UserWarning (integer regression targets get detected as classification — pass model= to be safe).
Output stays 3D (n_estimators, n_samples, embed_dim) — not a 2D drop-in for Pipeline / ColumnTransformer; README documents this

## Backward compatibility

`get_embeddings(...)`, `tabpfn_clf=`, and `tabpfn_reg=` still work and emit DeprecationWarning. Legacy get_embeddings preserves plain-KFold behavior for byte-compatible OOF numbers.

## Tests

New TestTabPFNEmbedding suite covering vanilla / K-fold for classifier and regressor, OOF vs final-model transform, fit_transform semantics, invalid n_fold, NotFittedError, shuffle decoupling, and auto-detect warnings.
Legacy tests preserved under TestTabPFNEmbeddingLegacy.


examples/embedding/README.md
```diff
@@ -1,17 +1,39 @@
 # TabPFN Embeddings
 
-The `TabPFNEmbedding` class is a utility for extracting embeddings from `TabPFNClassifier` or `TabPFNRegressor` models. It supports both standard training (vanilla embeddings) and K-fold cross-validation for embedding extraction. The embeddings can be used to enhance the performance of downstream machine learning models.
+`TabPFNEmbedding` is a scikit-learn style transformer that extracts embeddings from `TabPFNClassifier` or `TabPFNRegressor`. It supports both vanilla embeddings and K-fold cross-validated (out-of-fold) embeddings, which can be used to enhance the performance of downstream machine learning models.
```
adrian-prior: So, this is only scikit-learn "style" because the output format isn't correct? Or what is the difference between this and a "real" sckit-learn transformer?

examples/embedding/README.md
```diff
@@ -1,17 +1,39 @@
...
 
 ## Key Features
 
-- **Vanilla Embeddings**: Extract embeddings by training on the entire dataset when `n_fold=0`.
-- **K-Fold Cross-Validation**: Improve embedding effectiveness by using K-fold cross-validation when `n_fold>0`, as proposed in the paper *"A Closer Look at TabPFN v2: Strength, Limitation, and Extension"*.
+- **Vanilla embeddings**: with `n_fold=0`, the model is trained once on the entire training set and used to embed both train and test data.
```
adrian-prior, line 7: As far as I understand it, the authors of "A Closer Look at TabPFN v2" argue that there is a distribution shift between train and test embeddings because the train embeddings can attend to the label, while the test embeddings can't/attend to a dummy label. I think we can mention this here, i.e., we can quickly explain to the reader why we believe that the CV embeddings are good (and oc give credit to the authors)

examples/embedding/get_embeddings.py
```diff
@@ -18,76 +19,69 @@
 from tabpfn_extensions import TabPFNClassifier, TabPFNRegressor
 from tabpfn_extensions.embedding import TabPFNEmbedding
 
-# Load and evaluate classification dataset
-print("Loading classification dataset (kc1)...")
+# ---------------------------------------------------------------------------
+# Classification
+# ---------------------------------------------------------------------------
```
adrian-prior, line 32: I let CC run the examples on a GPU node, and the regression results look a bit weird to me. What do you think? Any idea why we are so much worse than the baseline?
Image

src/tabpfn_extensions/embedding/tabpfn_embedding.py
```diff
@@ -103,45 +262,45 @@ def get_embeddings(
...
+            "is deprecated and will be removed in a future release. Use "
+            "`embedding.fit_transform(X_train, y_train)` for OOF training "
+            "embeddings or `embedding.fit(X_train, y_train).transform(X)` for "
+            "unseen data. Note: the new `fit` path uses StratifiedKFold for "
+            "classifiers (this legacy path uses plain KFold), so OOF numbers "
+            "may differ.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        template = (
```
adrian-prior, line 250: I am in favor of just removing this if we think it's deprecated. If someone wants to use the old, they should just downgrade their extensions package, in my opinion.

src/tabpfn_extensions/embedding/tabpfn_embedding.py
```diff
@@ -1,100 +1,259 @@
...
+        """Return a fresh TabPFN model to use, warning on legacy kwargs."""
+        legacy = self.tabpfn_clf if self.tabpfn_clf is not None else self.tabpfn_reg
+        if legacy is not None:
+            warnings.warn(
+                "`tabpfn_clf` and `tabpfn_reg` are deprecated; pass `model=` "
+                "instead (or omit it to auto-detect classifier vs regressor "
+                "from y).",
+                DeprecationWarning,
+                stacklevel=3,
+            )
```
adrian-prior, line 115: If we do this, why don't we prevent the user from passing `tabpfn_cls`/`tabpfn_reg`. Shouldn't they just pass the task type and random state?

src/tabpfn_extensions/embedding/tabpfn_embedding.py
```diff
@@ -1,100 +1,259 @@
...
+        splits = (
+            cv.split(np.zeros(n), np.asarray(y)) if stratify else cv.split(np.zeros(n))
+        )
+
+        chunks: list[np.ndarray] = []
+        val_indices: list[np.ndarray] = []
+        for train_idx, val_idx in splits:
+            X_tr = _safe_row_index(X, train_idx)
+            y_tr = _safe_row_index(y, train_idx)
+            X_val = _safe_row_index(X, val_idx)
```
adrian-prior: Wouldn't it be easier to just cast everything to `np.ndarray` before the indexing/when calling _compute_oof`?

src/tabpfn_extensions/embedding/tabpfn_embedding.py
```diff
@@ -1,100 +1,259 @@
...
+        """Build the K-fold splitter used by both fit and the legacy path."""
+        rs = random_state if shuffle else None
+        if stratify:
+            return StratifiedKFold(
+                n_splits=self.n_fold,
+                shuffle=shuffle,
+                random_state=rs,
+            )
+        return KFold(n_splits=self.n_fold, shuffle=shuffle, random_state=rs)
+
```
adrian-prior: `_nit_`: I think this could be moved within the scope of `_compute_oof` as it's only used in there (can still be a separate function for readability).

tests/test_embedding.py
```diff
@@ -321,3 +322,226 @@ def test_embeddings_with_multiclass(self, dataset_generator):
...
+        """transform(X_train) goes through the final full-data model — it does
+        NOT return cached OOF embeddings, and should therefore differ from
+        train_embeddings_ when CV is on.
+        """
+        X_train, _X_test, y_train, _ = classification_data
+        clf = TabPFNClassifier(n_estimators=1, random_state=42)
+        transformer = TabPFNEmbedding(n_fold=3, model=clf)
+        transformer.fit(X_train, y_train)
+
+        oof = transformer.train_embeddings_
```
adrian-prior: `_nit_`: Would be cool to actually use fit_transform to get the train embeddings here.

tests/test_embedding.py
```diff
@@ -321,3 +322,226 @@ def test_embeddings_with_multiclass(self, dataset_generator):
...
+    def test_auto_task_classification_warns(
+        self,
+        classification_data: tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray],
+    ) -> None:
+        """Without an explicit model, classification is inferred and warns."""
+        X_train, _, y_train, _ = classification_data
+        transformer = TabPFNEmbedding(n_fold=0)
+        with pytest.warns(UserWarning, match="No `model=` provided"):
+            transformer.fit(X_train, y_train)
+        assert isinstance(transformer.model_, TabPFNClassifier)
```
adrian-prior, line 480: I think we should just force the user to set the target types and random state. I'm not a big fan of silently inferring the target type, even if we log a warning.

examples/embedding/get_embeddings.py
adrian-prior: It would also be cool, if we could reference this in the documentation as an example that using more folds actually improves performance -> i.e. it gives better embeddings. (After we know what is going on in regression).

adrian-prior (commented):
Thanks a lot for setting this up! Mostly looks good to me!

I left a bunch of comments. In general, I think we can reduce the amount of code we need to maintain by (a) directly removing deprecated code and (b) forcing the user to make some simple choices at init vs. trying to infer them from the input.

Let me know what you think!

examples/embedding/README.md
eliott-kalfon: The main difference with the scikit-learn style is the added dimension over the n_estimators

examples/embedding/README.md
eliott-kalfon: Sure, I also think we should push users to use the cross-validated embeddings (which is why we have default number of fold at 5). We should also do this in our [docs](https://docs.priorlabs.ai/capabilities/embeddings) that currently list the robust embeddings as the second option. Happy to mention the paper here too.

examples/embedding/get_embeddings.py
eliott-kalfon: I agree on this for the documentation. I will add the mention and review what you are showing for regression. That does look strange to me as well. I must say that I rarely use embeddings for regression tasks (where target encoding of categorical values generally performs very well)

src/tabpfn_extensions/embedding/tabpfn_embedding.py
eliott-kalfon: Fair, I like this stance. More direct

src/tabpfn_extensions/embedding/tabpfn_embedding.py
eliott-kalfon: They could also pass a model if they want. But I like the idea that using tabpfn_cls or reg can be removed straightaway. In the error message of trying to set it we could also show how to do it with the new version

tests/test_embedding.py
eliott-kalfon: Could do that. Important to note that this task type [inference](https://github.com/PriorLabs/tabpfn-extensions/pull/298/changes#diff-95bd9a2de74b0f0bb1f4d053e911d41994158cbaad41f40e81dd3dec6f91ec33:~:text=target_type%20%3D%20type_of_target(y)) works with the scikit-learn `type_of_target` method, a common practice

tests/test_embedding.py
eliott-kalfon: I think that this test tests the transform method. The `fit_transform` one is the one below

examples/embedding/get_embeddings.py
eliott-kalfon: I get the same when rerunning from `main`. This issue was there before. As I am not a fan of using embeddings for regression, I would be in favour of removing the regression code. That would also reduce the amount of code to maintain and review

adrian-prior: Hmm, I would also not really use the embeddings for predictive models, but rather for some other downstream task (like plotting, clustering, etc.). I think using them for regression/classification is just a simple proxy here for embedding quality (We could also just use TabPFN directly for classification and regression, which should yield better performance anyhow).

src/tabpfn_extensions/embedding/tabpfn_embedding.py
adrian-prior: Feel free to resolve the comment once this removed :)

brendan-priorlabs: The whole point of deprecation warnings is that we don't randomly break people when they upgrade the package. Deprecate first, remove later

src/tabpfn_extensions/embedding/tabpfn_embedding.py
adrian-prior: Sorry,
I did actually miss that. I think it's a good idea to ensure the user can pass a model. There is a slight difference in behavior here, isn't it? i.e., if I don't pass a model, then I can use it for both regression and classification. If I do pass a model, I can only use it for one of the two. I don't think we catch any missmatch between model type and task type in this case.

Did you check if the error message is informative?

adrian-prior: Regarding the warning: If we remove the parameters entirely, the users would just get an error and look at the docstring, which I think works just as well. Additionally, this is signature is entirely new, isn't it? I wouldn't be too worried about "backwards compatibility" in this case :)

src/tabpfn_extensions/embedding/tabpfn_embedding.py
```diff
@@ -1,100 +1,259 @@
...
+    return arr[idx]
+
+
+def _n_samples(X) -> int:
+    return X.shape[0] if hasattr(X, "shape") else len(X)
+
+
+def _validate_embedding_model(model) -> None:
+    """Raise if the model can't produce embeddings (e.g. TabPFN client)."""
+    if "tabpfn_client" in str(type(model).__module__):
```
brendan-priorlabs: We should just use `isinstance` on the class from the OSS package. Stringifying module names is rather brittle

src/tabpfn_extensions/embedding/tabpfn_embedding.py
```diff
@@ -1,100 +1,259 @@
...
+        raise ImportError(
+            "TabPFN embeddings require the full TabPFN implementation "
+            "(pip install tabpfn). The TabPFN client does not support "
+            "embedding extraction.",
+        )
+    if not hasattr(model, "get_embeddings"):
```
brendan-priorlabs: This shouldn't be necessary after the above type check

src/tabpfn_extensions/embedding/tabpfn_embedding.py
```diff
@@ -1,100 +1,259 @@
...
-        Args:
-            tabpfn_clf (Optional[TabPFNClassifier]): An instance of TabPFN classifier (if available).
-            tabpfn_reg (Optional[TabPFNRegressor]): An instance of TabPFN regressor (if available).
-            n_fold (int): Number of folds for cross-validation. If 0, cross-validation is not used.
+    def transform(self, X: np.ndarray) -> np.ndarray:
+        """Embed ``X`` using the fitted full-data model.
 
-        Raises:
-            ImportError: If using TabPFN client implementation which doesn't support embeddings.
+        This always uses ``model_`` — it does NOT return cached OOF
```
brendan-priorlabs: This comment is very focused on implementation. Can we reword it to be more focused on the user of this class? That is, explain to / remind them when to use `transform` vs. `fit_transform`

brendan-priorlabs (changes requested):
Thanks a lot for this change, @eliott-kalfon! Left a couple minor comments here (and one earlier on deprecation warnings -- I'd support them for this change [the overhead here seems fairly low]), but overall this looks quite nice :)

src/tabpfn_extensions/embedding/tabpfn_embedding.py
brendan-priorlabs: +1, a one-time conversion to `np.ndarray` makes sense to me. Let me know if we're missing something!

src/tabpfn_extensions/embedding/tabpfn_embedding.py
adrian-prior: Yeah, I agree that this is definitely the more principled approach. No strong opinions on this. We did take a bit of a hammer and ax approach to `tabpfn-extensions` previously, but now that the package is in better shape, this should probably change.

tests/test_embedding.py
adrian-prior: After our discussion this is ok by me :)

tests/test_embedding.py
adrian-prior: Yeah, what I wanted to say is that we could directly test that fit_transform does not give the same result as `fit` and then `transform` (`fit`/`transform`)

examples/embedding/README.md
adrian-prior: Sounds great!

src/tabpfn_extensions/embedding/tabpfn_embedding.py
brendan-priorlabs: IMO we can move fast and break things when needed. Here, however, the cost to doing backwards compat and potentially preserving some user goodwill is low, so it's worth doing that

tests/test_embedding.py
eliott-kalfon: Ok got it!

examples/embedding/get_embeddings.py
```diff
@@ -1,152 +1,138 @@
 """TabPFN Embedding Example
 
 This example demonstrates how to extract embeddings from TabPFN models and use them
-for classification and regression tasks.
+for classification and regression tasks via the scikit-learn style
+``TabPFNEmbedding`` transformer.
 
 NOTE: This example requires the full TabPFN implementation (pip install tabpfn).
 It will not work with the TabPFN client (pip install tabpfn-client) because
 the embedding functionality is not available in the client version.
```
adrian-prior, line 13: Could you also add a note here saying that this is just to demonstrate embedding quality? Embeddings will and probably should not be used to do what we do below in real-life.

examples/embedding/README.md
```diff
@@ -1,17 +1,39 @@
...
+- **K-fold cross-validation (recommended)**: with `n_fold>=2`, out-of-fold (OOF) embeddings are produced for the training data. Because TabPFN attends to training labels, train-set embeddings experience a distribution shift relative to test embeddings (where a dummy label is used). OOF embeddings fix this: each training sample is embedded in a fold where it was held out — matching the no-label-attention condition of unseen data. This is the robust variant described in *"A Closer Look at TabPFN v2: Strength, Limitation, and Extension"* (Ye et al., 2025, [arXiv:2502.17361](https://arxiv.org/abs/2502.17361)). A final model is also trained on the full training set for embedding new data.
```
adrian-prior: I think we can point to the example script to show that using hold-out embeddings for training data works better than using the default mode

src/tabpfn_extensions/embedding/tabpfn_embedding.py
adrian-prior: Reopening this comment: I still have a slight preference for dropping `tapfn_reg` and `tabpfn_clf`. As the signature of this class is new, I don't think we need to pruposfully reintroduce legacy code.

adrian-prior: (We can still and probably should keep the other legacy code around)

adrian-prior (approved):
LGTM! Left some minor comments!

src/tabpfn_extensions/embedding/tabpfn_embedding.py
eliott-kalfon: Ok, thanks for sharing! @brendan-priorlabs , do you have a view there?

We have two ways of doing this: keeping the args and deprecating, removing these arguments all together.

I like the idea of deprecation, but that is only my view

brendan-priorlabs: > As the signature of this class is new

What do you mean by this, @adrian-prior? (Typically we speak of method/function signatures, not class signatures.) This class existed before and continues to exist. If somebody upgrades the package, it will break them. Hence deprecation notice being preferred. Let me know if I'm missing something!

src/tabpfn_extensions/embedding/tabpfn_embedding.py
```diff
@@ -12,31 +12,24 @@
...
-
 def _validate_embedding_model(model) -> None:
     """Raise if the model can't produce embeddings (e.g. TabPFN client)."""
-    if "tabpfn_client" in str(type(model).__module__):
+    try:
+        from tabpfn import (
+            TabPFNClassifier as _LocalCLF,
+            TabPFNRegressor as _LocalREG,
+        )
+    except ImportError:
```
brendan-priorlabs: nit: Prefer top-level imports. And we don't need the `ImportError` checking as these are required anyway. LLMs like to do this overly defensive style of programming. IMO it makes things more brittle and should be avoided.

brendan-priorlabs (commented):
Small note here. Once we resolve the thread on the deprecation, should be good to go!

src/tabpfn_extensions/embedding/tabpfn_embedding.py
adrian-prior: My bad, I overlooked that the class already existed. I now see where the confusion comes from. Sorry about! Looks good from my side!

src/tabpfn_extensions/embedding/tabpfn_embedding.py
brendan-priorlabs: To be clear, I didn't mean that this entire validation function needed to be dropped. Just that the `except ImportError` wasn't helping. Sorry if that was confusing

brendan-priorlabs (approved):
LGTM! Thanks, @eliott-kalfon!

src/tabpfn_extensions/embedding/tabpfn_embedding.py
eliott-kalfon: That is my oversight, I read this way too quickly at the end of the day. Let me make sort this out and ask for review. Apologies for the extra work and thanks for the catch

PR #299, opened 2026-05-27
Added bar distribution documentation

eliott-kalfon:
Added to this repo following the Linear [ticket](https://linear.app/priorlabs/issue/PRI-54/document-bar-distribution-bimodal-full-uncertainty-regression-outputs). I would recommend adding this as an example script to the public repo going forward.


bejaeger (approved):
Nice write up!

Could you add a test to ensure `examples/predictive_distribution/predictive_distribution_example.py` works?
Ideally adding a CLI to the script so you can easily run it with a tiny dataset in the test so it's fast?

PR #300, opened 2026-05-28
chore: remove stale classifier_as_regressor references

adrian-prior:
## Summary

The `classifier_as_regressor` module was deleted in #35 (2025-03-21, commit `985dd3b`), but two stale references were left behind:

- `README.md` — the "Available Extensions" list still advertised it as a feature.
- `pyproject.toml` — the empty `classifier_as_regressor = []` install extra.

This PR removes both. No code changes; no impact on anything that imports from the package (since the module is already gone). The install extra was a no-op (empty list) so dropping it has no dependency effect.

Spotted while auditing extension coverage for [PRI-278](https://linear.app/priorlabs/issue/PRI-278/reference-all-the-sources-of-documentation-in-tabpfn-extensions).


oscarkey (approved):
lovely late spring cleaning!

PR #311, opened 2026-06-08
Fix outlier detection on MPS by casting target to logits dtype

adrian-prior:
## What

`TabPFNUnsupervisedModel.outliers` failed on the MPS (Apple Silicon) backend with:

```
TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64.
```

at `unsupervised.py:640`, reproducible by running `examples/unsupervised/density_estimation_outlier_detection.py` on a Mac.

## Root cause

The example passes `torch.tensor(X)` where `X` is sklearn float64, so the target `y_predict` is float64. The line moved it to the logits' device without casting:

```python
y_tensor = y_predict.clone().detach().to(logits.device)
```

MPS does not support float64, so the `.to(mps)` move raises. This is **not** a regression from the recent log-space outliers change (#291) — that PR only swapped `pdf`→`forward` on the lines below; the `.to(logits.device)` line predates it. CI runs on CPU, where float64 is fine, so it was never caught.

## Fix

Cast the target to the logits' dtype (float32) when moving it to the device. CPU behaviour is unchanged.

## Verification

- `examples/unsupervised/density_estimation_outlier_detection.py` now runs to completion on MPS, producing finite `log_p` of shape `(569,)`, range `[-8.02, 7.78]`.
- `examples/unsupervised/imputation.py` unaffected (already passing).
- Pinned `ruff==0.8.6` check + format and the full pre-commit suite pass.


src/tabpfn_extensions/unsupervised/unsupervised.py
```diff
@@ -637,7 +637,16 @@ def outliers_single_permutation_(
...
+                # Match logits dtype/device: MPS rejects float64, and sklearn
+                # inputs arrive as float64, so cast before moving to the device.
+                y_tensor = (
+                    y_predict.clone()
+                    .detach()
+                    .to(
+                        dtype=logits.dtype,
+                        device=logits.device,
+                    )
+                )
```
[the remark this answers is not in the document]
adrian-prior: Good catch — done in bd5dbc7. Since `.to()` already returns a new tensor when the dtype/device changes and the result is only read by `criterion.forward`, the `.clone()` was a redundant copy. Removed.

adrian-prior:
@eliott-kalfon I assigned you to the PR because @anuragg1209 is on holiday for a few weeks. Hope that is ok :)

src/tabpfn_extensions/unsupervised/unsupervised.py
```diff
@@ -637,7 +637,12 @@ def outliers_single_permutation_(
                 pred = model.predict(X_predict, output_type="full")
                 logits = pred["logits"]
                 logits_tensor = logits.clone().detach()
-                y_tensor = y_predict.clone().detach().to(logits.device)
+                # Match logits dtype/device: MPS rejects float64, and sklearn
```
eliott-kalfon, line 756: nit: do we generally leave these comments?

eliott-kalfon (approved):
Hi @adrian-prior, thanks for the fix. I agree, it was broken and it now works.

On this PR: Does the example also run with CUDA after this change?
Future consideration (separation of concerns): Should this example output something to make this feature more approachable?

src/tabpfn_extensions/unsupervised/unsupervised.py
adrian-prior: In this case, I would leave it just so people don't change it back in the future. WDYT?

adrian-prior:
@eliott-kalfon Thanks for the review! I will verify that this actually still runs on CUDA, but it definitely should.

Re: Example: I am unsure if I understand what you are saying. Would you prefer that users get the normal vs. the log probs in the example?

adrian-prior:
@eliott-kalfon I tested it on a CUDA gpu and it still runs -> Will merge.

PR #312, opened 2026-06-08
Share TabPFN class-count inference; separate categorical detection from class capacity

### LINKED ISSUE #304: TabPFNUnsupervisedModel caps categorical features at 10 classes, ignoring TabPFN v3's 160-class capacity
### Describe the bug

TabPFNUnsupervisedModel (unsupervised extension: outlier detection, imputation, synthetic data)
only models a categorical feature with the classifier when it has fewer than 10 distinct values —
even on TabPFN v3, whose classifier supports 160 classes. Categorical features with 10–160 distinct
values are silently treated as numerical and modeled by the regressor instead.

There are two independent causes:

1. `use_classifier_` (src/tabpfn_extensions/unsupervised/unsupervised.py) reads
   `getattr(self.tabpfn_clf, "max_num_classes_", 10)`. TabPFN estimators do not expose a
   `max_num_classes_` attribute (neither before nor after `fit`), so the fallback of 10 always
   applies. This is the same dead lookup that #282 removed from `ManyClassClassifier`, where it was
   replaced by reading `get_inference_config().MAX_NUMBER_OF_CLASSES`.

2. `infer_categorical_features` (src/tabpfn_extensions/utils.py, called by
   `TabPFNUnsupervisedModel.fit`) hard-codes `max_unique_values_as_categorical_feature = 10`, so a
   feature the user explicitly declared categorical is silently dropped from
   `categorical_features` during `fit` if it has more than 10 distinct values. This gate runs
   first, so fixing (1) alone has no effect for declared categoricals.

Both gates should honor the classifier's real capacity (160 on v3, with a fallback of 10 for older
models / backends without `get_inference_config()`), following the #282 precedent.

There is also a small inconsistency between the two gates today: retention during `fit` is
inclusive (`n_unique <= 10`) while routing is exclusive (`n_unique < 10`), so a feature with
exactly 10 distinct values is kept as "categorical" yet still routed to the regressor.

I have a fix ready following the #282 pattern (resolve the capacity once from
`get_inference_config().MAX_NUMBER_OF_CLASSES`, fallback 10; make the two gates consistent;
backward-compatible parameter on `infer_categorical_features` so its other caller,
post_hoc_ensembles, is unaffected) and will open a PR referencing this issue.

### Steps/Code to Reproduce

```python
import numpy as np
from tabpfn import TabPFNClassifier, TabPFNRegressor
from tabpfn_extensions.unsupervised import TabPFNUnsupervisedModel

# TabPFN v3 supports 160 classes:
print("model capacity:", TabPFNClassifier().get_inference_config().MAX_NUMBER_OF_CLASSES)

# A column declared categorical with 12 distinct values (well within 160)
X = np.column_stack([
    np.repeat(np.arange(12), 5),                 # categorical: 12 distinct values
    np.random.default_rng(0).normal(size=60),    # numerical
]).astype(np.float32)

model = TabPFNUnsupervisedModel(tabpfn_clf=TabPFNClassifier(), tabpfn_reg=TabPFNRegressor())
model.set_categorical_features([0])
model.fit(X)

print("categorical_features after fit:", model.categorical_features)
print("routed to classifier:", model.use_classifier_(0, X[:, 0]))
```

### Expected Results

The declared categorical column (12 distinct values, far below the model's 160-class capacity)
is retained as categorical and modeled by the classifier:

model capacity: 160
categorical_features after fit: [0]
routed to classifier: True

### Actual Results

```
The column is silently dropped from categorical_features during fit and routed to the regressor:

model capacity: 160
categorical_features after fit: []
routed to classifier: False

No error or warning is raised. The dead attribute is verifiable directly:

>>> hasattr(TabPFNClassifier(), "max_num_classes_")
False
```

### Versions

```shell
PyTorch version: 2.12.0
OS: macOS 15.3 (arm64)
Python version: 3.11.13
Is CUDA available: False
CPU: Apple M1

Dependency Versions:
--------------------
tabpfn: 8.0.4
torch: 2.12.0
numpy: 2.4.6
scipy: 1.17.1
pandas: 3.0.3
scikit-learn: 1.9.0
typing_extensions: 4.15.0
einops: 0.8.2
huggingface-hub: 1.17.0

tabpfn-extensions: 0.4.2 (current main, 37ae137)
```

adrian-prior:
> **Credit:** This overlaps with #305 by @ClementBourt, which first identified the hard-coded class limit and proposed the shared `get_max_num_classes` utility (in a TODO) that this PR implements. Credited as co-author on all commits. Opened #305 before this — see the discussion there on how to reconcile the two.

## Problem

The number of classes TabPFN can predict was **hard-coded to 10** in several places. With TabPFN-3 the real limit is much higher (`get_max_num_classes(TabPFNClassifier()) == 160`), so `TabPFNUnsupervisedModel.use_classifier_` (which read a non-existent `max_num_classes_` and always fell back to 10) wrongly routed categorical columns with >10 levels to the regressor, and `infer_categorical_features` conflated "is this column categorical?" with "can the model predict it?".

## Changes

**1. Single source of truth — `utils.get_max_num_classes(model)`.** Extracted from `ManyClassClassifier._get_alphabet_size` (preserving its logic): reads `get_inference_config().MAX_NUMBER_OF_CLASSES`, falls back to 10 for older TabPFN, returns `None` for non-TabPFN estimators. `many_class` and `unsupervised` both call it, so one fix propagates everywhere.

**2. Separate the two constraints** for treating a column as categorical:
- **(a) actually categorical** — `infer_categorical_features` now answers only this: respects caller-declared categoricals unconditionally, and auto-detects low-cardinality columns up to `MAX_UNIQUE_VALUES_FOR_CATEGORICAL` (**10 → 40**), gated by a samples-per-category ratio (`> MIN_SAMPLES_PER_CATEGORY = 10`, replacing the flat `>100` rows guard).
- **(b) model can predict it** — applied separately at the point of use (`use_classifier_` checks `n_unique <= get_max_num_classes(clf)`).

**3. Fail fast on a missing model.** `TabPFNUnsupervisedModel` still allows one of `tabpfn_clf`/`tabpfn_reg` to be `None`, but `__init__` now `raise`s (not `assert`, which `python -O` strips) when both are None, and `density_` raises a clear `ValueError` naming the column that needs the absent estimator — instead of a cryptic `AttributeError` on `None`. `use_classifier_` uses `torch.unique` (GPU-safe) instead of `np.unique`.

## Behavior changes (intentional)

- Auto-detection now flags columns with up to 40 unique values (was <10), requires >10 samples/category, and keeps caller-declared categoricals regardless of cardinality.
- `use_classifier_` uses `<=` against the model's real class limit (was `< 10`).

## Verification

- `get_max_num_classes`: 160 for `TabPFNClassifier`, `None` for `DecisionTreeClassifier`.
- Missing-model and both-None configs raise clear `ValueError`s.
- `tests/test_many_class_classifier.py` + `tests/test_unsupervised.py` green; pinned ruff + full pre-commit (incl. mypy) pass.

Fixes #304


src/tabpfn_extensions/many_class/many_class_classifier.py
```diff
@@ -108,20 +109,8 @@ def _resolve_row_weighting_config(
...
-        # Older TabPFN versions (pre PriorLabs/TabPFN#890) don't expose
-        # get_inference_config; their MAX_NUMBER_OF_CLASSES was always 10.
-        if type(self.estimator).__module__.startswith("tabpfn"):
-            logger.info(
-                "Base estimator has no get_inference_config(); assuming "
-                "alphabet_size=10 (older TabPFN default).",
-            )
-            return 10
-        return None
+        # Shared inference so a fix to the class-count logic applies everywhere.
```
oscarkey: don't need this comment?

src/tabpfn_extensions/utils.py
```diff
@@ -393,32 +393,86 @@ def get_tabpfn_models() -> tuple[type, type]:
...
+    """
+    if hasattr(model, "get_inference_config"):
+        cfg = model.get_inference_config()
+        val = getattr(cfg, "MAX_NUMBER_OF_CLASSES", None)
+        if val:
+            return int(val)
+    # Config-less TabPFN builds were always capped at 10; non-TabPFN estimators
+    # have no inherent limit, so signal that with None.
+    if type(model).__module__.startswith("tabpfn"):
+        return DEFAULT_MAX_NUM_CLASSES
```
oscarkey: Should we just bump the minimum tabpfn version to 8.0.0. Then we don't need the defaults?

oscarkey (changes requested):
lgtm! just two questions

src/tabpfn_extensions/utils.py
adrian-prior: Sounds good, will do.

src/tabpfn_extensions/many_class/many_class_classifier.py
adrian-prior: Removed in f63b7f7.

src/tabpfn_extensions/utils.py
adrian-prior: Done in f374398 — bumped the floor to `tabpfn>=8.0.0`, removed `DEFAULT_MAX_NUM_CLASSES` and the config-less fallback branch in `get_max_num_classes`, and reworked `use_classifier_` so it no longer needs a default. Added a `breaking` changelog fragment for the min-version bump.

oscarkey (approved):
sweeeet

PR #316, opened 2026-06-08
fix: bind client TabPFN wrapper classes so the API backend resolves

### LINKED ISSUE #315: get_tabpfn_models() raises NameError when the API backend is selected (USE_TABPFN_LOCAL=false)
### Describe the bug

Importing `tabpfn_extensions` (or `tabpfn_extensions.utils`) crashes with a
NameError whenever the API/client backend is selected via
`USE_TABPFN_LOCAL=false` and `tabpfn-client` is installed.

`utils.py` resolves the backend eagerly at import time with a module-level
call `TabPFNClassifier, TabPFNRegressor = get_tabpfn_models()`. In the
client branch, `get_tabpfn_models()` reads the names `TabPFNClassifierWrapper`
/ `TabPFNRegressorWrapper`, but those names are only ever assigned in the
`except ImportError:` fallback (set to `None`). On the success path — where
`tabpfn-client` imports fine and the wrapper classes `ClientTabPFNClassifier`
/ `ClientTabPFNRegressor` are defined — the `...Wrapper` names are never bound,
so the `elif TabPFNClassifierWrapper is not None:` line dereferences an
undefined name.

Because the failing call sits at module top level, this is an *import-time*
crash: every `from tabpfn_extensions ... import ...` fails before any model is
constructed and before any API call is attempted. The local backend
(`USE_TABPFN_LOCAL=true`, the default) is unaffected because it returns earlier
and never reads the undefined names — which is why the bug is invisible in the
default configuration and in any test run with local weights.

I have the fix ready locally i will open a PR once i published this issue.

### Steps/Code to Reproduce

```python
# Requires: pip install tabpfn-client   (the client must import successfully)
import os

# Select the API/client backend. utils.py reads this at import time.
os.environ["USE_TABPFN_LOCAL"] = "false"

from tabpfn_extensions.utils import get_tabpfn_models  # <-- raises at import
```

### Expected Results

The module imports successfully and `get_tabpfn_models()` returns the two
client wrapper classes (ClientTabPFNClassifier, ClientTabPFNRegressor), the
same way it returns (LocalTabPFNClassifier, LocalTabPFNRegressor) when
USE_TABPFN_LOCAL=true. No error is thrown.

### Actual Results

```python-traceback
Traceback (most recent call last):
  File "repro.py", line 6, in <module>
    from tabpfn_extensions.utils import get_tabpfn_models
  File ".../tabpfn_extensions/utils.py", line 393, in <module>
    TabPFNClassifier, TabPFNRegressor = get_tabpfn_models()
                                        ^^^^^^^^^^^^^^^^^^^
  File ".../tabpfn_extensions/utils.py", line 383, in get_tabpfn_models
    elif TabPFNClassifierWrapper is not None:
         ^^^^^^^^^^^^^^^^^^^^^^^
NameError: name 'TabPFNClassifierWrapper' is not defined
```

### Versions

```shell
PyTorch version: 2.11.0
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 15.3 (arm64)
Clang version: 14.0.3 (clang-1403.0.22.14.1)

Python version: 3.11.13 (main, Aug 18 2025, 19:05:01) [Clang 20.1.4 ] (64-bit runtime)
Python platform: macOS-15.3-arm64-arm-64bit
Is CUDA available: False
Is XNNPACK available: True

CPU:
Apple M1

Dependency Versions:
--------------------
tabpfn: 7.1.1
torch: 2.11.0
numpy: 2.3.5
scipy: 1.16.3
pandas: 2.3.3
scikit-learn: 1.7.2
typing_extensions: 4.15.0
einops: 0.8.2
huggingface-hub: 1.14.0
```

ClementBourt:
## Summary

Selecting the API/client backend with `USE_TABPFN_LOCAL=false` (with
`tabpfn-client` installed) crashes at **import time** with:

    NameError: name 'TabPFNClassifierWrapper' is not defined

`utils.py` resolves the backend eagerly via a module-level
`TabPFNClassifier, TabPFNRegressor = get_tabpfn_models()`. In the client
branch the selector reads `TabPFNClassifierWrapper` / `TabPFNRegressorWrapper`,
but those names are only assigned (to `None`) in the `except ImportError:`
fallback. On the success path the wrapper classes are defined as
`ClientTabPFNClassifier` / `ClientTabPFNRegressor`, so the names the selector
reads are never bound — and because the call runs at module top level, the
whole package fails to import.

## Root cause

The success path and the selector disagree on the identifier:

- `try:` block defines `class ClientTabPFNClassifier(...)` /
  `class ClientTabPFNRegressor(...)`.
- `except ImportError:` sets `TabPFNClassifierWrapper = None` /
  `TabPFNRegressorWrapper = None`.
- `get_tabpfn_models()` reads `TabPFNClassifierWrapper` in its `elif` and
  returns the `...Wrapper` pair.

With `tabpfn-client` importable, the `except` branch never runs, so the
`...Wrapper` names are undefined when the `elif` dereferences them.

Note: the local backend (`USE_TABPFN_LOCAL=true`, the default) returns before
the `elif`, so it never touches the undefined names — which is why the bug is
invisible in the default config and in local-only test runs.

## Fix

Use one consistent set of identifiers on both paths: rename
`TabPFNClassifierWrapper` → `ClientTabPFNClassifier` and
`TabPFNRegressorWrapper` → `ClientTabPFNRegressor` in all four occurrences
(the two `except`-branch `None` sentinels and the two reads in
`get_tabpfn_models`). After the change the name is always bound — to the class
on success, to `None` on failure — and the selector returns the client wrapper
classes correctly.

Closes #315


adrian-prior (commented):
Thanks a lot for setting this up! Could you (a) set up some tests to make sure that a similar regression doesn't happen again, and (b) also cover some of the other edge-cases?

ClementBourt:
On it.

adrian-prior:
@ClementBourt Thanks a lot! Just rerequest my review once you added the tests :)

ClementBourt:
While crafting the test i realized that there was an asymmetry in backend selection guarding. The commit message sums it up quite well imo but basicly I figured that there is a more conservative approach to backend selection.
Whenever the backend selection could not be local there was a client fallback. I have few reason to believe this is worth reconsidering :

1. Client is not part of the base dependencies in .toml. 
2. If a user hasn't explicitly mentioned they want to use the client then making them use it will induce unwanted API cost on their side and might hide the fact that the local setup they required is buggy. This could also make them fall back to a default model selection without realizing. I don't see why they would want to use a version earlier than version 3 but i guess the world is creative when it comes to unforeseen situations. 

I took the liberty to change the backend selection such that it strictly follow the client parameters avoiding any surprises. I do not claim to be right on this at all so i will happily roll back these changes if need be. 
I added the test that both cover this new logic and guard against the previous regression.

changelog/315.fixed.md
```diff
@@ -0,0 +1 @@
+Fix `get_tabpfn_models()` raising `NameError` at import time when the API/client backend is selected (`USE_TABPFN_LOCAL=false`) with `tabpfn-client` installed: the client wrapper classes are now bound to the names the backend selector reads, so API inference resolves correctly again.
```
adrian-prior: I think this isn't up-to-date anymore given the larger scope. Could you update this?

tests/test_backend_selection.py
```diff
@@ -0,0 +1,121 @@
...
+) -> None:
+    """get_tabpfn_models returns the selected backend, or raises if it is missing."""
+    _set_backends(
+        monkeypatch,
+        use_local=use_local,
+        local_present=local_present,
+        client_present=client_present,
+    )
+
+    if expected == "error":
```
adrian-prior, line 110: _nit_: Maybe we can also assert here that we give the user the correct error message?

adrian-prior (approved):
Great spot and execution! Thanks a lot for this! All the comments I left are minor so feel free to resolve and merge once you addressed them :) (or just resolve them if you think they aren't applicable)

tests/test_backend_selection.py
ClementBourt: added two test for this.

changelog/315.fixed.md
ClementBourt: Done !

ClementBourt:
Thank you for the feedbacks ! As you suggested i have resolved your comments after addressing them.

adrian-prior:
@ClementBourt next time you can also merge the PR yourself, if you want :)

PR #321, opened 2026-06-23
test: run example-file tests in pull request CI

### LINKED ISSUE #51: Improve testing: reenable testing example files
Due to an issue properly handling example file timeouts (they are very slow to run on CPU) the example tests are currently disabled. We'd like to reenable, starting to run example files for 5 seconds, which will catch some of the errors in there, then aborting with XFAIL. We have this implemented and executed if pytest is run with --run_examples but by default this is disabled.

mvanhorn:
## Summary
The `tests/test_examples.py` harness (5s timeout + XFAIL for slow examples, gated behind `--run-examples`) is implemented, but the CI pytest steps ran `pytest tests/` without `--run-examples`, so example files were never imported in CI — which is exactly what #51 asks for.

## Changes
Add `--run-examples` to the Linux and Windows pytest steps in `.github/workflows/pull_request.yml`, plus a changelog entry.

Fixes #51

AI was used for assistance.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -113,13 +113,13 @@ jobs:
         env:
           TABPFN_EXCLUDE_DEVICES: mps
         run: |
-          FAST_TEST_MODE=1 uv run --no-sync pytest tests/
+          FAST_TEST_MODE=1 uv run --no-sync pytest tests/ --run-examples
 
       - name: Run Tests (Windows)
         if: runner.os == 'Windows'
         run: |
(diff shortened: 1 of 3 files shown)


adrian-prior:
Thanks for kicking this off, @mvanhorn 🙏 — closing in favour of #327, which carries it forward.

#327 keeps the same goal (run the example files in CI) but swaps the bounded-timeout/XFAIL harness for one that fails on real errors, and fixes the example breakages it surfaced instead of XFAILing past them. Your work here is credited as co-author on #327.

PR #325, opened 2026-06-25
Speed up test suite ~1.6x in fast mode (318s -> 195s)

LeoGrin:
## Summary

Reduces per-test cost in `FAST_TEST_MODE`, cutting the suite from **318s → 195s (~1.6×)** measured serially on a 4-core machine. Tests still run **serially** (no pytest-xdist). Both changes are scoped to fast mode, so non-fast / local behaviour is unchanged, and the pass set is identical before/after.

1. **`tests/conftest.py`** — the shared `tabpfn_classifier` / `tabpfn_regressor` fixtures now use `n_estimators=1` in fast mode (was TabPFN's default of 8). That's ~8× fewer forward passes on every test built on these fixtures (base, dt_pfn, rf_pfn, embedding). Ensembling is still exercised explicitly by `test_ensemble_configurations`. Non-fast mode still uses 8.

2. **`tests/test_pval_crt.py`** — the two slow *active* CRT tests cost ~90s combined. In fast mode they use fewer permutations `B` and smaller datasets, preserving each test's intent:
   - `test_crt_output_contract` only checks output structure → `B=40→5`, `n=120→60`.
   - `test_crt_detects_obvious_signal` relies on the p-value floor `1/(B+1)` → `B=40→30` keeps it at ~0.032, safely under the `< 0.05` assertion; `n=200→120`.

## Measurements (4-core, `FAST_TEST_MODE=1`, serial)

| Run | pytest time | speedup |
|---|---|---|
| Baseline | 318.3s | 1.0× |
| This PR | 195.3s | 1.6× |

Identical pass set: 102 passed / 44 skipped (plus 11 pre-existing failures in the local test env that lacked the optional `autogluon` dep — unrelated to this PR).

## Notes
- Biggest remaining serial cost is `test_crt_detects_obvious_signal` (~29s); reducing it further would need `tabpfn_crt` to expose `n_estimators` (a library change), so it's left as a follow-up.

---

> [!NOTE]
> **Low Risk**
> Test-only changes gated on FAST_TEST_MODE with ensemble coverage retained in dedicated tests; no production code paths affected.
> 
> **Overview**
> When **`FAST_TEST_MODE=1`**, shared **`tabpfn_classifier`** / **`tabpfn_regressor`** fixtures now construct estimators with **`n_estimators=1`** via **`FAST_FIXTURE_KWARGS`**, cutting ensemble work across most TabPFN extension tests. Outside fast mode the fixtures pass no override so backend defaults stay the same; **`test_ensemble_configurations`** still sweeps **`n_estimators`** in `[1, 5]`.
> 
> Regressor ensemble tests were updated from the old **`N_ensemble_configurations`** attribute to **`n_estimators`**, matching the current estimator API.
> 
> **`test_pval_crt.py`** mirrors fast-mode sizing: structure-only **`test_crt_output_contract`** uses smaller **`n`** and **`B`**, while **`test_crt_detects_obvious_signal`** keeps **`B`** high enough for the **`p_value < 0.05`** bound under the **`1/(B+1)`** floor.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f1ad08b5679e7a136360687cd8b1b627d3ee8f27. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tests/test_pval_crt.py
```diff
@@ -76,10 +83,12 @@ def test_crt_output_contract():
 @pytest.mark.local_compatible
 def test_crt_detects_obvious_signal():
     """Non-flaky: signal is so strong that failure is astronomically unlikely."""
+    n = 120 if FAST_TEST_MODE else 200
     rng = np.random.RandomState(0)
-    X = rng.randn(200, 4)
+    X = rng.randn(n, 4)
     y = 100 * X[:, 1]  # noiseless, massive coefficient
 
+    # B must stay >=40 so the p-value floor 2/(B+1) is < 0.05.
```
psinger-prior: If not mistaken, based on https://github.com/PriorLabs/tabpfn-extensions/blob/main/src/tabpfn_extensions/pval_crt/crt.py#L423 the floor should be 1/(B+2)

The PR description also says: test_crt_detects_obvious_signal relies on the p-value floor 1/(B+1) → B=40→30 keeps it at ~0.032, safely under the < 0.05 assertion; n=200→120.

So should this be changed to 30?

psinger-prior (changes requested):
I noted unrelated to this PR that the regressor test might be using a wrong `n_estimator` setting (probably outdated): https://github.com/PriorLabs/tabpfn-extensions/blob/main/tests/test_base_tabpfn.py#L451

tests/test_pval_crt.py
LeoGrin: Thanks! Both addressed in f1ad08b:

- The floor is `1/(B+1)` — `crt.py:423` divides by `B+1` — so fast-mode `B=30` gives `1/31 ≈ 0.032 < 0.05`. I'd accidentally left the code at `B=40` while the description still said 30; it's back to `B = 30 if FAST_TEST_MODE else 40` so the code and description match.
- Good catch on the regressor `test_ensemble_configurations`: it was setting the outdated `N_ensemble_configurations` (silently ignored), so the `[1, 5]` loop wasn't testing anything. It now sets `n_estimators`, and I renamed the loop variable to match.

adrian-prior:
@psinger-prior Could you give the PR another pass?

psinger-prior (approved):

PR #326, opened 2026-06-25
Fix unsupervised experiments marking every feature categorical

### LINKED ISSUE #323: GenerateSyntheticDataExperiment appears to treat every feature as categorical
### Describe the bug

I am trying to use TabPFN to generate synthetic data so that I can include it as a baseline in my paper.

However, there is no option in the GenerateSyntheticDataExperiment for the user to select the categorical features. It seems in the existing implementation that the categorical features are just set to all the selected indices? https://github.com/PriorLabs/tabpfn-extensions/blob/38c6561395559ffaf913b5f956012aaf4c31a2e9/src/tabpfn_extensions/unsupervised/experiments.py#L124-L131

In the meantime, I have gotten around this for my own implementation by patching the run method to allow the user to pass their own selected categorical features in the kwargs.

### Steps/Code to Reproduce

```python
import numpy as np
from tabpfn import TabPFNClassifier, TabPFNRegressor
from tabpfn_extensions.unsupervised import TabPFNUnsupervisedModel
from tabpfn_extensions.unsupervised.experiments import GenerateSyntheticDataExperiment

X = np.column_stack([
    np.repeat(np.arange(15), 4),     # categorical: 15 distinct values
    np.random.default_rng(0).normal(size=60),    # numerical
]).astype(np.float32)

y = np.array([])
feature_names = ["categorical", "numerical"]
feature_indices = [*range(X.shape[1])]

model_unsupervised = TabPFNUnsupervisedModel(
    tabpfn_clf=TabPFNClassifier(), 
    tabpfn_reg=TabPFNRegressor()
)

# Run synthetic data generation experiment
experiment = GenerateSyntheticDataExperiment(
    task_type="unsupervised"
)

results = experiment.run(
    tabpfn=model_unsupervised,
    X=X,
    y=y,
    attribute_names=feature_names,
    temp=1.0,                     # Temperature parameter for sampling
    n_samples=X.shape[0],     # Generate as many samples as original data
    indices=feature_indices,
)

print("categorical_features after fit:", model_unsupervised.categorical_features)

data = experiment.data
real_data = synth_data = data[data["real_or_synthetic"] == "Actual samples"]
synth_data = data[data["real_or_synthetic"] == "Generated samples"]

print("Real data:")
print(real_data.head())
print("Synthetic data:")
print(synth_data.head())
```

### Expected Results

Numerical features to be generated with float (rather than integer) data.

### Actual Results

```
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:04<00:00,  2.37s/it]
categorical_features after fit: [0, 1]
Real data:
   categorical  numerical real_or_synthetic
0          0.0   0.125730    Actual samples
1          0.0  -0.132105    Actual samples
2          0.0   0.640423    Actual samples
3          0.0   0.104900    Actual samples
4          1.0  -0.535669    Actual samples
Synthetic data:
   categorical  numerical  real_or_synthetic
0          9.0        3.0  Generated samples
1          2.0        2.0  Generated samples
2          4.0        2.0  Generated samples
3         13.0        2.0  Generated samples
4         12.0        2.0  Generated samples
```

### Versions

```shell
PyTorch version: 2.11.0+cu128
OS: Ubuntu 22.04.5 LTS (x86_64)
Python version: 3.10.0 (default, Mar  3 2022, 09:58:08) [GCC 7.5.0] (64-bit runtime)
Is CUDA available: True
GPU models and configuration: 
GPU 0: NVIDIA RTX A6000
Nvidia driver version: 570.211.01

Dependency Versions:
--------------------
tabpfn: 8.0.8
torch: 2.11.0+cu128
numpy: 2.2.6
scipy: 1.15.3
pandas: 2.3.3
scikit-learn: 1.7.2
typing_extensions: 4.15.0
einops: 0.8.2
huggingface-hub: 1.20.1


tabpfn-extensions: (I installed the current main from github, 38c6561)
```

alanprior:
GenerateSyntheticDataExperiment and OutlierDetectionUnsupervisedExperiment computed categorical indices with a comprehension that always evaluated to [0, 1, ..., n-1], so every selected feature was force-marked categorical and numerical columns were generated as integers (issue #323). There was also no way for a caller to specify which features are categorical.

Replace the broken comprehension in both experiments with a `categorical_features` kwarg (indices in the original X column space, same as `indices`), translated to the selected-subset position. When omitted it defaults to [] so the model's existing infer_categorical_features auto-detection runs at fit time. Also add a `should_plot=True` guard to GenerateSyntheticDataExperiment so data-only/headless callers can skip the forced plot, and document the kwargs.

Add tests/test_unsupervised_experiments.py covering the index translation, the default-empty behavior, and an end-to-end check that numerical columns are generated as floats.

Fixes #323

### DIFF
--- a/tests/test_unsupervised_experiments.py
+++ b/tests/test_unsupervised_experiments.py
@@ -136,6 +136,33 @@ def test_generate_synthetic_experiment_translates_original_indices_to_subset():
 
 @pytest.mark.client_compatible
 @pytest.mark.local_compatible
+def test_generate_synthetic_experiment_accepts_array_indices():
+    """Array-typed indices and categorical_features are accepted, not just lists."""
+    X = np.column_stack(
+        [
+            np.random.default_rng(0).normal(size=120),
+            np.repeat(np.arange(4), 30),
+        ],
+    ).astype(np.float32)
+
+    model = _RecordingUnsupervisedModel()
+    experiment = GenerateSyntheticDataExperiment(task_type="unsupervised")
+    experiment.run(
+        tabpfn=model,
+        X=X,
+        y=np.array([]),
+        attribute_names=["numerical", "categorical"],
+        indices=np.array([0, 1]),
+        categorical_features=np.array([1]),
+        n_samples=10,
+        should_plot=False,
+    )
+
+    assert model.categorical_features == [1]
+
+
+@pytest.mark.client_compatible
+@pytest.mark.local_compatible
 def test_outlier_experiment_uses_supplied_categorical_features():
     """OutlierDetection shares the same fix: respect the supplied categorical list."""
     X = torch.tensor(
--- a/tests/test_unsupervised_experiments.py
+++ b/tests/test_unsupervised_experiments.py
@@ -1,0 +1,206 @@
+"""Tests for the unsupervised experiment wrappers in ``experiments.py``.
+
+These cover the categorical-feature handling of ``GenerateSyntheticDataExperiment``
+and ``OutlierDetectionUnsupervisedExperiment`` (regression test for issue #323,
+where every selected feature was force-marked categorical and numerical columns
+were generated as integers).
+
+The fast tests use a lightweight recording stand-in so they need no TabPFN
+weights; one end-to-end test uses a real model under ``FAST_TEST_MODE=1``.
+"""
+
+from __future__ import annotations
+
+import numpy as np
+import pytest
+import torch
+
+from tabpfn_extensions import TabPFNClassifier, TabPFNRegressor, unsupervised
+
+# experiments.py imports matplotlib/seaborn at import time; skip the whole module
+# if those optional plotting dependencies are not installed.
+experiments = pytest.importorskip("tabpfn_extensions.unsupervised.experiments")
+GenerateSyntheticDataExperiment = experiments.GenerateSyntheticDataExperiment
+OutlierDetectionUnsupervisedExperiment = (
+    experiments.OutlierDetectionUnsupervisedExperiment
+)
+
+
+class _RecordingUnsupervisedModel:
+    """Stand-in for ``TabPFNUnsupervisedModel`` that needs no TabPFN weights.
+
+    It records the categorical-feature indices the experiment passes to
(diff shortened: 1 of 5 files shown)


tests/test_unsupervised_experiments.py
```diff
@@ -0,0 +1,231 @@
...
+        return np.random.default_rng(0).normal(size=(n_samples, n_features))
+
+    def outliers(self, X, n_permutations=3):
+        return torch.zeros(X.shape[0], dtype=torch.float32)
+
+
+@pytest.mark.client_compatible
+@pytest.mark.local_compatible
+def test_generate_synthetic_experiment_uses_supplied_categorical_features():
+    """Only the caller-designated column is marked categorical (issue #323)."""
```
adrian-prior: _nit_: I don't think you need to mention the issue here.

tests/test_unsupervised_experiments.py
```diff
@@ -0,0 +1,231 @@
...
+    )
+
+    assert model.categorical_features == [0]
+    assert "log_p" in result
+
+
+@pytest.mark.client_compatible
+@pytest.mark.local_compatible
+def test_generate_synthetic_experiment_keeps_numerical_features_numerical(monkeypatch):
+    """End-to-end: a numerical column must not be generated as integers (issue #323)."""
```
adrian-prior: Same here.

adrian-prior (approved):
Looks good to me! Thanks a lot for looking into this!

Before you merge this (if you haven't already done so), actually look at how the generated categorical columns look for some example dataset (e.g., via plotting). Do they look similar to the input dataset? E.g., are we generating categories that weren't in the input dataset? Does the distribution look similar, etc.? This "test" could even become part of the example script.

(and you still need to add a changelog ;))

tests/test_unsupervised_experiments.py
alanprior: ofc :\ thanks for that

alanprior:
Thanks @adrian-prior , great comment

Issue #225, opened 2025-12-12
TabEBM broken with recent TabPFN + test broken.


YFrisch:
### Describe the bug

`from tabpfn_extensions.tabebm.tabebm import TabEBM`

throws

```
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[16], line 1
----> 1 from tabpfn_extensions.tabebm.tabebm import TabEBM
      3 print("\n" + "=" * 50)
      4 print("GENERATING SYNTHETIC DATA WITH TabEBM")

File ~/.../site-packages/tabpfn_extensions/tabebm/tabebm.py:16
     13 from tabpfn_common_utils.telemetry import set_extension
     14 from torch.utils.data import DataLoader
---> 16 from tabpfn.config import ModelInterfaceConfig, PreprocessorConfig
     17 from tabpfn.utils import meta_dataset_collator
     18 from tabpfn_extensions.utils import TabPFNClassifier

ModuleNotFoundError: No module named 'tabpfn.config'
```

tabpfn==6.0.6
tabpfn-extensions @ git+https://github.com/PriorLabs/tabpfn-extensions.git@7a900321c6aaea8e58ae23117628eaf18cc5a4ab

### Versions

```shell

```

oscarkey:
hey @YFrisch , I noticed you closed this but as far as I can tell it's still an issue?

It seems like it wasn't picked up because the tests catch import error and use this to decide not to test TabEBM: https://github.com/PriorLabs/tabpfn-extensions/blob/1960cc63a419f9022e902c17bb2c5407ed9e5431/tests/test_tabebm.py#L21
so we should fix that too.

PR #327, opened 2026-06-25
test: run example files as CI tests; fix bugs they surface (PRI-328)

### LINKED ISSUE #51: Improve testing: reenable testing example files

adrian-prior:
Closes PriorLabs/tabpfn-extensions#51.

## What

Reenables example-file testing in CI, replacing the old 5s-timeout/XFAIL smoke harness with a real one.

* **Harness rewrite** (`tests/test_examples.py`): each example runs as a subprocess. Two modes:
  * *Smoke* (per-PR, default): short timeout; an example that **errors** fails, one that **times out** passes (we only assert it starts and runs without crashing).
  * *Strict* (`--example-strict`, for the scheduled GPU run): a timeout **is** a failure — run to completion.
* **CPU smoke workflow** (`.github/workflows/example_tests.yml`): path-filtered to `examples/**` + the harness; runs the smoke gate with `-n auto`.
* **Bug fixes** — running the examples for real surfaced these:
  * shapiq examples → bump `interpretability` floor to `shapiq>=1.2.0` (lowest-direct was installing 1.1.0, which predates `class_index` (1.1.1) and `TabPFNExplainer` (1.2.0)).
  * `shap_example` → `shap` added to a new `examples` dependency group, floored `>=0.46.0` (0.41.0 pulled an ancient, unbuildable numba).
  * `generate_data_following_dag` → import the `unsupervised.experiments` submodule explicitly.
  * `tabebm` example → xfail + documented; broken with current TabPFN's `inference_config` API (PriorLabs/tabpfn-extensions#225).

## Verified

Built the env on a GPU node and ran the examples — `shapiq_example` + `generate_data_following_dag` pass, `tabebm` xfails. Full CPU smoke suite: **16 passed, 2 skipped** (`survival`: GPL; `get_embeddings`: GPU-only), **1 xfailed**.

## Follow-up

Nightly GPU run-to-completion (strict) + making slow examples test-mode-aware is tracked in [PRI-330](https://linear.app/priorlabs/issue/PRI-330/nightly-gpu-example-tests-run-all-examples-to-completion-strict-make).

## Notes

Supersedes PriorLabs/tabpfn-extensions#321 (the prior attempt at PriorLabs/tabpfn-extensions#51) — builds on @mvanhorn's work there, credited as co-author.


.github/workflows/example_tests.yml
```diff
@@ -0,0 +1,79 @@
+name: Example smoke tests
+run-name: "Example smoke on ${{ github.head_ref || github.ref_name }}"
+
+# Fast per-PR smoke gate for the scripts in examples/. Each example is run as a
+# subprocess; the test fails only if it *errors*. An example that does not finish
+# within the timeout still passes -- we only assert it starts and runs without
+# crashing. Running every example to completion (where a timeout is a failure) is
+# the job of the scheduled GPU run; see PRI-330.
+#
+# Path-filtered so this only runs when an example or the harness changes (~most
```
oscarkey: shouldn't we also run it when the extensions themselves change?

tests/test_examples.py
```diff
@@ -1,228 +1,232 @@
...
-        timeout: Optional timeout parameter (no longer used internally, kept for backward compatibility)
-
-    Returns:
-        The imported module object
-    """
-    # Add the parent directory to sys.path to allow imports within example files
-    parent_dir = str(path.parent)
-    if parent_dir not in sys.path:
-        sys.path.insert(0, parent_dir)
+def _cuda_available() -> bool:
```
oscarkey: don't think we need this function, can just call `torch.cuda.is_available()`?

.github/workflows/example_tests.yml
oscarkey, line 1: what's the rationale for separating this from the standard test workflow?

tests/test_examples.py
```diff
@@ -1,228 +1,232 @@
...
+        timeout = request.config.getoption("--example-long-timeout")
+    else:
+        timeout = request.config.getoption("--example-timeout")
+
+    # Examples are top-to-bottom scripts; run each in its own process so a hang
+    # can be killed cleanly and state never leaks between examples. The example
+    # inherits TEST_MODE/FAST_TEST_MODE/TABPFN_EXCLUDE_DEVICES from this process.
+    env = dict(os.environ)
+    env["TEST_MODE"] = "1"
+    # Prepend the in-repo src/ so the example subprocess imports this checkout's
```
oscarkey: Is there a particular reason we do this in this case? It's more normal for the tests to assume the environment is set up correctly.

pyproject.toml
```diff
@@ -101,6 +103,20 @@ dev = [
...
+# not need these, only the example demos do, and the examples ship in the repo (not
+# the wheel), so anyone running them works from a clone where this group is available.
+examples = [
+    # interpretability/shap_example.py bridges shapiq values into a shap.Explanation
+    # purely for plotting (shap.summary_plot / shap.plots.*). The library dropped shap
+    # as a runtime dep in #283 (RES-1467) since shapiq supersedes the adapter; it
+    # survives only as an example plotting convenience.
+    # Floor is 0.46.0, not the older 0.41.0: 0.41.0 pins an ancient numba (0.47.0)
+    # with no cp310+ wheel, which fails to build from source. 0.46.0 leaves numba
+    # unconstrained so a modern, wheel-backed numba is used.
```
oscarkey: I think this bit of the comment is more detail than we need

pyproject.toml
```diff
@@ -56,10 +56,12 @@ source = "https://github.com/PriorLabs/tabpfn-extensions"
...
 interpretability = [
-    # 1.1.0 is the lowest release where `imputer="baseline"` + `index="SV"` +
-    # `max_order=1` (the default config used by get_tabpfn_imputation_explainer)
-    # works without raising.
-    "shapiq>=1.1.0",
+    # 1.2.0 is the lowest release providing every shapiq feature this package
+    # uses: `class_index` on TabularExplainer/Explainer (added in 1.1.1) and
+    # `shapiq.TabPFNExplainer` (added in 1.2.0, used by get_tabpfn_explainer).
+    # With the old `>=1.1.0` floor, CI's lowest-direct resolution installed
+    # 1.1.0 and the examples crashed (missing class_index / TabPFNExplainer).
```
oscarkey: I don't think we need this comment? It's rare to reduce the minimum version, and people would expect that doing this might break things.

oscarkey (changes requested):
thanks for doing this! Seems like a very good idea. Just a few questions…

.github/workflows/example_tests.yml
adrian-prior: Good call — folded it into `pull_request.yml` as an always-run `example_smoke` job and deleted the separate file. The only reason it was separate was path-filtering, and once it should also run on `src/` changes (your next comment) that filter excluded almost nothing. Done in e7178e8.

.github/workflows/example_tests.yml
adrian-prior: Agreed — it now runs on every PR with no path filter, so library changes are covered too. Done in e7178e8.

tests/test_examples.py
adrian-prior: Done — inlined `torch.cuda.is_available()` and dropped the wrapper.

tests/test_examples.py
adrian-prior: Removed — the subprocess now relies on the installed package like the rest of the tests, no PYTHONPATH manipulation. (It was originally added on a Gemini suggestion; agreed your convention is cleaner.)

pyproject.toml
adrian-prior: Trimmed to a one-liner.

pyproject.toml
adrian-prior: Trimmed.

tests/test_examples.py
```diff
@@ -1,227 +1,222 @@
...
+    if example_file["long"]:
+        timeout = request.config.getoption("--example-long-timeout")
+    else:
+        timeout = request.config.getoption("--example-timeout")
+
+    # Examples are top-to-bottom scripts; run each in its own process so a hang
+    # can be killed cleanly and state never leaks between examples. The example
+    # inherits TEST_MODE/FAST_TEST_MODE/TABPFN_EXCLUDE_DEVICES from this process.
+    env = dict(os.environ)
+    env["TEST_MODE"] = "1"
```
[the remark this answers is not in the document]
adrian-prior: Reverting this per maintainer review — we rely on the editable install (`uv sync`) rather than manipulating PYTHONPATH.

tests/test_examples.py
```diff
@@ -1,227 +1,222 @@
...
+            f"(smoke mode; completion not verified)",
+        )
+        return
+
+    if proc.returncode != 0:
+        output = (proc.stdout or b"").decode("utf-8", "replace")
+        tail = "\n".join(output.strip().splitlines()[-25:])
+        pytest.fail(
+            f"Example {name} exited with code {proc.returncode}:\n{tail}",
+        )
```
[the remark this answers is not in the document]
adrian-prior: Kept — failure output now shows the last 100 lines.

adrian-prior:
@oscarkey Thanks for all the comments! This wasn't my best PR, to be honest. Should be better now. Can you give it another pass?

.github/workflows/pull_request.yml
```diff
@@ -202,6 +202,60 @@ jobs:
           key: model-cache-${{ hashFiles('model_cache/**') }}
           enableCrossOsArchive: true
 
+  example_smoke:
```
oscarkey, line 212: Would you add a comment that we use a separate job for the examples, so that they run in parallel without blocking the normal tests (as they're slow)

tests/conftest.py
```diff
@@ -121,6 +121,26 @@ def pytest_addoption(parser):
...
+        default=120,
+        help="Per-example timeout in seconds for normal examples (see tests/test_examples.py)",
+    )
+    parser.addoption(
+        "--example-long-timeout",
+        action="store",
+        type=int,
+        default=120,
+        help="Per-example timeout in seconds for long-running examples (a timeout never fails these)",
+    )
```
oscarkey: maybe we just hard-code these timeout options in `test_examples.py`? We don't really want the args to pytest controlling whether tests pass or fail.

pyproject.toml
```diff
@@ -114,6 +122,9 @@ log_cli = false
 log_level = "DEBUG"
 xfail_strict = true
 addopts = "--durations=10 -vv"
+markers = [
+    "example: runs an example script from examples/ (deselect with -m 'not example')",
+]
```
oscarkey: We don't seem to be using the marker for anything? So we could get rid of it?

tests/test_examples.py
```diff
@@ -1,227 +1,213 @@
...
+"""Run the example files in ``examples/`` and check they work.
+
+Each example is executed as a subprocess. The outcome is interpreted with one
+of two policies, selected per CI layer:
+
+* **Per-PR smoke (default):** each example gets a short timeout. An example that
+  *errors* fails the test; one that simply doesn't finish in time *passes* — we
+  only assert it starts and runs without crashing. This is a fast, cheap guard
+  against broken example scripts and import errors.
+* **Weekly full run (``--example-strict``):** a longer timeout, and a normal
```
oscarkey: Maybe "Regularly scheduled full run". Otherwise this is period to go out of sync with whatever action is doing the scheduling

oscarkey (approved):
nice!

.github/workflows/pull_request.yml
adrian-prior: Done in 98b86d2 — added a comment on the job explaining it's separate so the (slower) examples run in parallel without blocking the normal test suite.

tests/conftest.py
adrian-prior: Agreed. Hard-coded the smoke timeout as a constant in test_examples.py and dropped the --example-timeout/--example-strict/--example-long-timeout options entirely. They were only there for the scheduled full run (strict mode), which isn't in this PR — that's PRI-330, which will add strict mode along with the nightly job. Done in 98b86d2.

pyproject.toml
adrian-prior: Good catch — nothing selected on it (examples are gated by --run-examples, not -m example). Removed the marker and its registration in 98b86d2.

tests/test_examples.py
adrian-prior: Done — reworded to 'regularly scheduled full run' in 98b86d2.

PR #335, opened 2026-07-03
add simple imputation method (PRI-313)

jmkuebler:
## Summary
The current imputation method [`TabPFNUnsupervisedModel.impute`](https://github.com/PriorLabs/tabpfn-extensions/blob/8d11f8b8b04c0e92071e3399538255fa05074188/src/tabpfn_extensions/unsupervised/unsupervised.py#L648) is relatively involved and slow. 

This PR adds a simple lightweight alternative that simply imputes column-by-column by predicting the missing values from all the rows that contain values for the present columns. Missing values in the remaining columns are treated natively by TabPFN. It always conditions on all other features.

So far we extended the example to just include it on top of the existing.

**Discussion points**:

- [x] Shall we keep the simple method lightweight (no DAG support, no support for ordering, no temperature sampling)? If a user needs them they can implement it, but otherwise it inflates the method. I'd propose to keep it lean.
- [x] Shall we remove the old method or keep it? @noahho 
- [ ] Do we need more testing? Or any special dataset to run this through?
- [ ] So far only tested on CPU. Is that fine?

### Benefits
**Speed**: The method is faster than the existing method. The new method predicts each missing column **exactly once**. The old method iterates over columns to impute, but in each iteration temporarily imputes all the other columns first (and then discards those).   For a single permutation that is `k + (k−1) + … + 1 = k(k+1)/2` predict calls (previously-imputed columns are skipped in later iterations), and this is repeated for each of the n_permutations (default `10`) permutations.

**Accuracy**: The simple method can be more accurate because it can condition on more rows, namely all that have a value for the present column. The existing method instead has a fixed reference set that is used for all columns.

### Testing
We tested on four `sklearn` datasets (on CPU) and swiped the number of estimators (for the simple method) and the number of permutations for the existing method. On those cases the simple method is consistently faster and better (for reasons outlined above).
benchmark_plot

### DIFF
--- a/examples/unsupervised/imputation.py
+++ b/examples/unsupervised/imputation.py
@@ -75,3 +75,25 @@ original_values = X_test[original_nan_mask]
 
 mse = np.mean((imputed_values - original_values) ** 2)
 print(f"Mean Squared Error of imputed values vs. original values: {mse:.4f}")
+
+# --- 6. Alternative: the simple, transductive imputer ---
+# `simple_impute` is a lightweight alternative that needs no separate fit step.
+# For each column with missing values it fits a TabPFN model on the rows where
+# that column is observed (using all other columns as features) and predicts the
+# missing rows. Here we hand it the training rows stacked on top of the test rows
+# so it has the (complete) training data as additional context.
+# thus when predicting each column it can have a larger context and thus
+# higher accuracy
+
+from tabpfn_extensions.unsupervised import simple_impute
+
+print("\nImputing the same data with simple_impute...")
+X_merged_missing = np.vstack([X_train, X_test_missing])
+X_merged_imputed = simple_impute(X_merged_missing, tabpfn_reg=reg, tabpfn_clf=clf)
+
+# Recover the imputed test rows and score them the same way as above.
+X_test_imputed_simple = X_merged_imputed[len(X_train) :]
+imputed_values_simple = X_test_imputed_simple[original_nan_mask]
+
+mse_simple = np.mean((imputed_values_simple - original_values) ** 2)
+print(f"[simple_impute] Mean Squared Error vs. original values: {mse_simple:.4f}")
--- a/examples/unsupervised/imputation.py
+++ b/examples/unsupervised/imputation.py
@@ -80,7 +80,7 @@ model_unsupervised.fit(X_train)
 print("Imputing missing values by sampling from the model...")
 X_imputed_tensor = model_unsupervised.impute(
     X_test_missing,
-    n_permutations=5,  # Fewer permutations for a quicker example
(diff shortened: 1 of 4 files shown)


noahho:
ping @adrian-prior

src/tabpfn_extensions/unsupervised/simple_impute.py
```diff
@@ -0,0 +1,137 @@
...
+        tabpfn_clf: TabPFN classifier used for categorical columns. Defaults to
+            ``TabPFNClassifier()`` when needed.
+        categorical_features: Indices of categorical columns. If None, they are
+            inferred from the data.
+
+    Returns:
+        np.ndarray: A new array with all missing values imputed.
+    """
+    X = np.array(X, dtype=np.float32, copy=True)
+    n_features = X.shape[1]
```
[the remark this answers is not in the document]
adrian-prior: I don't know about asserting that this has 2 dims, but checking that we have at least two columns seems reasonable doesn't it?

examples/unsupervised/imputation.py
```diff
@@ -36,42 +37,55 @@
...
-print("Fitting the unsupervised model on the training data...")
+print("\nFitting TabPFNUnsupervisedModel on the training data...")
 model_unsupervised.fit(X_train)
 
-# Perform imputation on the test set
-print("Imputing missing values...")
+print("Imputing missing values by sampling from the model...")
 X_imputed_tensor = model_unsupervised.impute(
     X_test_missing,
     n_permutations=5,  # Fewer permutations for a quicker example
```
adrian-prior: I think `n_permutations` isn't really needed here is it? Each estimator of TabPFN already permuates the column order (i.e. different feature ordering). This only makes sense to use when `use_imputed=True`, doesn't it?

adrian-prior (approved):
Sorry for the late review! I was OOO for the past two weeks!

Thanks for this PR! Only left some minor comments! Consider also adding a short page for missing value imputation to our docs :)

examples/unsupervised/imputation.py
jmkuebler: Just to be sure I get it right. You mean since the default model (used to build the UnsupervisedModel) uses 8 estimators by default, we should not have explicit permutations here?

I do agree. Note though that this is a code I didn't touch. But if we agree it should be 1, I\ll do that.

src/tabpfn_extensions/unsupervised/simple_impute.py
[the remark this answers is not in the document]
jmkuebler: sounds good, I added that.

examples/unsupervised/imputation.py
jmkuebler: Changed to 1

PR #338, opened 2026-07-07
Fix GenerateSyntheticDataExperiment doc string.

oscarkey:
Fixes PRI-334

Noah I'm not sure what the original intent of this code was, maybe you know?


noahho (approved):

noahho:
@oscarkey this is basically an example script in the form of a class. We would evaluate some of these experiments while running training.

PR #343, opened 2026-07-14
Add Bayesian optimization example with differentiable EI

klemens-floege:
## What

Adds `examples/bayesian_optimization/bayesian_optimization.py`: a self-contained Bayesian optimization loop using `TabPFNRegressor` as the surrogate model, in the spirit of [PFNs4BO](<https://github.com/automl/PFNs4BO>) (Müller et al., ICML 2023) but powered by the TabPFN foundation model.

Addresses PriorLabs/TabPFN#1116, which asks how to compute Expected Improvement from the predicted distribution and backpropagate to the candidate inputs. Originally proposed as PriorLabs/TabPFN#1118; moved here since it is an application example rather than core-package functionality.

The benchmark objective (Hartmann-6) comes from `botorch.test_functions`, so the script contains only the BO loop and the TabPFN surrogate code.

## How it works

Each BO iteration:

1. Fits on the evaluated points via `fit_with_differentiable_input` (tensors in, graph preserved).
2. Screens 512 random candidates with closed-form EI from `raw_space_bardist_.ei()` in a single forward pass — no Gaussian assumption, EI is integrated directly over the predicted bar distribution.
3. Refines the top 4 candidates by **gradient ascent on EI itself**: `differentiable_input=True` lets gradients flow from the acquisition value back to the candidate coordinates.
4. Evaluates the objective at the best candidate and repeats.

## Requirements / test integration

* Needs the full TabPFN package at **>= 8.1.0** (`fit_with_differentiable_input` on the regressor shipped in 8.1.0); does not work with the TabPFN client. The example directory is registered in `REQUIRES_TABPFN_DIRS` in `tests/test_examples.py` accordingly.
* Needs `botorch` for the benchmark objective (noted in the module docstring, not added to package dependencies).
* Follows the examples convention: no `main()` wrapper, so the example test harness picks it up (as a slow example it xfails on the 5 s timeout).

## Results

image


pyproject.toml
```diff
@@ -86,11 +86,17 @@ hpo = [
 survival = []
 many_class = []
 unsupervised = []
+bayesian_optimization = [
+    # fit_with_differentiable_input on the regressor shipped in tabpfn 8.1.0.
+    "tabpfn>=8.1.0",
```
eliott-kalfon: this is already required by the package core deps, I do not see the need of pinning there

eliott-kalfon (approved):
Hi Klemens, looks good to me! On another note, this is great to see BO in action with TabPFN
Another example we could add is showing how we can extract the gradient with regards to data. This could even go in the core package imo, as a super simple script

klemens-floege:
> Another example we could add is showing how we can extract the gradient with regards to data. This could even go in the core package imo, as a super simple script

Let me get on that!

pyproject.toml
klemens-floege: Done!

Issue #342, opened 2026-07-14
Switching to ProxySHAP/RegressionMSR as the default approximator for SHAP-value computation


adrian-prior:
Hi @Advueu963,

Thanks a lot for chatting at ICML. According to your suggestion, I am looking into switching our default approximator for SHAP-values in TabPFN from KernelSHAP to ProxySHAP. Before I can do that, I have a few questions:

- (1) Most of our users use first-order SHAP value computation vs. second and third-order. If I understand it correctly, for first-order SHAP values, ProxySHAP reduces to RegressionMSR (https://arxiv.org/pdf/2506.11849), so that is why there are no first-order results in the paper. Is that correct?
- (2) Is ProxySHAP/RegressionMSR universally a better choice than KernelSHAP? E.g., looking at Figure 2 in https://arxiv.org/pdf/2506.11849, it seems that only for a relatively large number of evaluated coalitions does RegressionMSR dominate KernelSHAP. If so, is there a heuristic we could use to decide when to switch from KernelSHAP to ProxySHAP? Does this heuristic change based on the order?

Looking forward to getting ProxySHAP into `tabpfn-extensions`!

Best,
Adrian

Advueu963:
Heyyy @adrian-prior,
Thanks for picking up our discussion from ICML! I hope you had a safe trip back :).

Yes, it is correct that ProxySHAP is identical to RegressionMSR for order-1 computation, as reflected in the shapiq codebase.
Regarding the second question, I would like to give a more detailed assessment, also considering the newest approximator available in shapiq:

**TL;DR:** There is also the ODDSHAP estimator, often beating RegressionMSR, so it might be interesting to have a routing to ODDSHAP for SHAP computation. For interactions, it is best to use ProxySHAP. The figures show Shapley Approximator overviews and RegressionMSR performance in lower-budget regimes.

1. **SHAP Computation [SHAP Ranking: [ODDSHAP](https://arxiv.org/pdf/2602.01399) > [RegressionMSR](https://arxiv.org/pdf/2506.11849) > [LeverageSHAP](https://arxiv.org/pdf/2410.01917) > KernelSHAP]**:
[...]

PR #345, opened 2026-07-15
Route TabPFN SHAP explainers to OddSHAP and ProxySHAP

adrian-prior:
## What

The shapiq interpretability wrappers (`get_tabpfn_imputation_explainer`, `get_tabpfn_explainer`) never set `approximator`, so shapiq's `"auto"` selection kicked in — **KernelSHAP** for first-order Shapley values and **KernelSHAPIQ** for interactions. Neither is the current best estimator.

This PR routes to the estimators recommended for TabPFN in the discussion at PriorLabs/tabpfn-extensions#342:

| Request | Approximator | Backend |
| -- | -- | -- |
| First-order Shapley values (`index="SV"`) | **OddSHAP** | LightGBM |
| Interaction indices (e.g. `k-SII`) | **ProxySHAP** | XGBoost |

Routing keys on `index` (`"SV"` → OddSHAP, which is Shapley-only; everything else → ProxySHAP, which supports all cardinal-probabilistic indices). A new explicit `approximator` argument on both wrappers lets callers override the routing (e.g. `approximator="auto"` for the old behavior).

## Dependencies

* Bump `shapiq>=1.1.0` → `shapiq>=1.6.0` (first release exposing OddSHAP/ProxySHAP).
* Add `xgboost` + `lightgbm` to the `interpretability` extra — the surrogate backends OddSHAP/ProxySHAP fit. Without them both silently fall back to a weaker `DecisionTreeRegressor`.

## Follow-up

shapiq's own `"auto"` does not yet select these estimators. Once it does ([mmschlk/shapiq#565](<https://github.com/mmschlk/shapiq/issues/565>)), this routing can be dropped in favor of `approximator="auto"`. Both the module docstring and `_resolve_approximator` note this.

## Testing

* Unit: `_resolve_approximator` routes `SV`→OddSHAP, `k-SII`→ProxySHAP, and passes explicit overrides through unchanged.
* End-to-end: both wrappers build explainers whose `.approximator` is the routed type and `.explain()` returns finite values on a real TabPFN model (synthetic data + the `examples/interpretability/shapiq_example.py` California-housing example, which computes first- and second-order values).
* Confirmed OddSHAP/ProxySHAP use their intended LightGBM/XGBoost backends (no DecisionTree fallback).
* `pre-commit run` (ruff, ruff-format, mypy) passes on the changed files.

Closes [RES-2045](https://linear.app/priorlabs/issue/RES-2045/are-we-using-proxyshap-in-tabpfn-extensions-interpretability).


src/tabpfn_extensions/interpretability/shapiq.py
```diff
@@ -71,6 +81,53 @@ def _build_tabular_explainer(shapiq, **kwargs):
...
+    """
+    if approximator is not None:
+        return approximator
+    # shapiq < 1.6.0 ships OddSHAP/ProxySHAP as ``None`` placeholders (the
+    # attribute exists but is not a class), so test the value, not ``hasattr``.
+    odd_shap = getattr(shapiq, "OddSHAP", None)
+    proxy_shap = getattr(shapiq, "ProxySHAP", None)
+    if index == "SV" and odd_shap is not None:
+        return odd_shap(n=n_features)
+    if index != "SV" and proxy_shap is not None:
```
eliott-kalfon: Super niche, Claude tells me that the "STII" index would break with proxyshap and give a key error. This may be too much of an edge case

eliott-kalfon:
```python
import warnings

import shapiq
from shapiq.explainer.configuration import setup_approximator

# --- verbatim copy of _resolve_approximator from
# src/tabpfn_extensions/interpretability/shapiq.py (PR #345, lines 84-128) ---


def _resolve_approximator(shapiq, approximator, index: str, n_features: int):
    if approximator is not None:
        return approximator
    odd_shap = getattr(shapiq, "OddSHAP", None)
    proxy_shap = getattr(shapiq, "ProxySHAP", None)
    if index == "SV" and odd_shap is not None:
        return odd_shap(n=n_features)
    if index != "SV" and proxy_shap is not None:
        return "proxyshap"
    version = getattr(shapiq, "__version__", None) or "< 1.6.0"
    warnings.warn(
        f"shapiq {version} does not provide the OddSHAP/ProxySHAP estimators "
        "recommended for TabPFN, so approximator='auto' "
        "(KernelSHAP/KernelSHAPIQ) is used instead. Upgrading to Python >= 3.12 "
        "and shapiq >= 1.6.0 enables the more accurate estimators.",
        stacklevel=2,
    )
    return "auto"


# --- end verbatim copy ---

N_FEATURES = 5

print(f"shapiq version: {shapiq.__version__}\n")

# ---------------------------------------------------------------------------
# Bug 1: index="STII" is a documented valid index for these explainers
# (shapiq.explainer.custom_types.ExplainerIndices includes it), but
# ProxySHAP's config doesn't support it -> raw KeyError instead of a
# graceful fallback/clear error.
# ---------------------------------------------------------------------------
print("=== Bug 1: index='STII' ===")
resolved = _resolve_approximator(shapiq, None, "STII", N_FEATURES)
print(f"_resolve_approximator returned: {resolved!r}  (looks fine so far)")

try:
    # This is exactly what shapiq.TabularExplainer.__init__ does internally
    # with the value _resolve_approximator just handed it.
    setup_approximator(resolved, index="STII", max_order=2, n_players=N_FEATURES, random_state=None)
except KeyError as e:
    print(f"BUG REPRODUCED: KeyError({e}) raised from inside shapiq's "
          f"setup_approximator -- get_tabpfn_explainer(..., index='STII') "
          f"would crash the same way.\n")
```

src/tabpfn_extensions/interpretability/shapiq.py
```diff
@@ -71,6 +81,53 @@ def _build_tabular_explainer(shapiq, **kwargs):
         return shapiq.TabularExplainer(**kwargs)
 
 
+def _resolve_approximator(shapiq, approximator, index: str, n_features: int):
```
eliott-kalfon: Another niche Claude but important: we may want to pass a random_state as argument to be passed as kwarg

src/tabpfn_extensions/interpretability/shapiq.py
```diff
@@ -71,6 +81,53 @@ def _build_tabular_explainer(shapiq, **kwargs):
...
+    full control.
+    """
+    if approximator is not None:
+        return approximator
+    # shapiq < 1.6.0 ships OddSHAP/ProxySHAP as ``None`` placeholders (the
+    # attribute exists but is not a class), so test the value, not ``hasattr``.
+    odd_shap = getattr(shapiq, "OddSHAP", None)
+    proxy_shap = getattr(shapiq, "ProxySHAP", None)
+    if index == "SV" and odd_shap is not None:
+        return odd_shap(n=n_features)
```
eliott-kalfon: random_state that we could pass here for reproducibility

eliott-kalfon (approved):
Solid PR thank you! I did some tests on a single dataset, comparing KernelSHAP + OddSHAP to exact SHAP values. OddSHAP was much better.
Did the same for KernelSHAPIQ and ProxySHAP, there the results were mixed. With ProxySHAP taking longer. This is only a sample of one.

Left some minor comments that Claude spotted.

src/tabpfn_extensions/interpretability/shapiq.py
adrian-prior: Added it

src/tabpfn_extensions/interpretability/shapiq.py
adrian-prior: Added it

src/tabpfn_extensions/interpretability/shapiq.py
adrian-prior: Should be fixed now :) (tested it using your script)

PR #348, opened 2026-07-20
Require DAGs for generation/imputation to specify every feature

adrian-prior:
## What

`generate_synthetic_data(dag=...)` and `impute(dag=...)` now raise a `ValueError` when the `dag` dict does not contain every feature index as a key. Parentless features must be stated explicitly with an empty list (`{i: []}`); the error message says exactly which features are missing and how to fix it.

## Why

Previously a partial DAG silently gave every unlisted feature an empty parent list, so it was generated **marginally** — independent of all other features. A user passing `dag={5: [2, 3]}` on a 20-feature dataset almost certainly means "constrain feature 5, stay agnostic about the rest", but there is no well-defined joint distribution for that reading — so instead of silently distorting the output, we fail loudly and make the user state their intent.

No deprecation period: the silent behavior is a correctness footgun, and spelling out the full DAG is a one-liner (`{**{i: [] for i in range(n)}, 5: [2, 3]}`).

## Changes

- `_resolve_dag_order` (`unsupervised.py`): new validation raising `ValueError` listing missing features; docstring updated.
- Docstrings of `impute_`, `impute`, `generate_synthetic_data`: state the full-specification requirement and the empty-list idiom.
- Tests: `test_resolve_dag_order_partial` → `test_resolve_dag_order_rejects_partial_dag` (rejection + explicit-roots acceptance); mutation test updated to a full DAG.
- The `generate_data_following_dag.py` example already specifies all 13 features — unchanged.
- `changelog/348.breaking.md`: towncrier news fragment (Breaking Changes) — the CI changelog gate checks for a fragment file, not direct `CHANGELOG.md` edits.

The corresponding `capabilities/data-generation.mdx` update is folded into PriorLabs/docs#81 (commit caca979), which should merge only after this PR is merged and released.

## Test plan

- [x] `FAST_TEST_MODE=1 pytest tests/test_unsupervised_dag.py` — 8 passed
- [x] `uvx pre-commit run` on changed files — clean


alanprior (approved):
LGTM

PR #354, opened 2026-07-21
fix: preserve client criterion in output_type="full" instead of importing removed tabpfn module

### LINKED ISSUE #350: Client full-output regression imports a removed TabPFN module
### Describe the bug

With the `tabpfn-client` backend, `ClientTabPFNRegressor.predict(..., output_type="full")` imports `tabpfn.model.bar_distribution.FullSupportBarDistribution` before delegating to the client.

That module no longer exists in the required `tabpfn>=8.1`. Current `tabpfn-client` already includes a compatible `criterion` in its full output, but the extension wrapper raises before using it.

### Steps/Code to Reproduce

```python
import os

os.environ["USE_TABPFN_LOCAL"] = "false"

import numpy as np
from tabpfn_extensions import TabPFNRegressor, unsupervised

X_train = np.linspace(-1.0, 1.0, 32, dtype=np.float32).reshape(-1, 1)
X_test = np.linspace(-1.5, 1.5, 16, dtype=np.float32).reshape(-1, 1)

model = unsupervised.TabPFNUnsupervisedModel(
    tabpfn_clf=None,
    tabpfn_reg=TabPFNRegressor(n_estimators=1, random_state=0),
).fit(X_train)

model.outliers(X_test, n_permutations=1)
```

This assumes `TABPFN_TOKEN` is configured.

### Expected Results

The wrapper should preserve the `criterion` returned by current `tabpfn-client`, and `outliers` should return one finite score per input row.

### Actual Results

```python-traceback
Found existing access token, reusing it for authentication.
00:10 Fitting... Done!
Traceback (most recent call last):
  File "/Users/missbrenner/src/tabpfn-extensions/src/tabpfn_extensions/utils.py", line 318, in predict
    from tabpfn.model.bar_distribution import FullSupportBarDistribution
ModuleNotFoundError: No module named 'tabpfn.model.bar_distribution'; 'tabpfn.model' is not a package

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/missbrenner/src/tabpfn-extensions/scripts/test.py", line 16, in <module>
    model.outliers(X_test, n_permutations=1)
  File "/Users/missbrenner/src/tabpfn-extensions/.venv/lib/python3.12/site-packages/tabpfn_common_utils/telemetry/core/decorators.py", line 218, in wrapped
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/Users/missbrenner/src/tabpfn-extensions/src/tabpfn_extensions/unsupervised/unsupervised.py", line 866, in outliers
    log_p = self.outliers_single_permutation_(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/missbrenner/src/tabpfn-extensions/src/tabpfn_extensions/unsupervised/unsupervised.py", line 753, in outliers_single_permutation_
    pred = model.predict(X_predict, output_type="full")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/missbrenner/src/tabpfn-extensions/src/tabpfn_extensions/utils.py", line 337, in predict
    raise ValueError(
ValueError: output_type='full' requires the TabPFN package with FullSupportBarDistribution to be installed
```

### Versions

```shell
Collecting system and dependency information...
PyTorch version: 2.12.1
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 26.5.2 (arm64)
GCC version: Could not collect
Clang version: 16.0.6
CMake version: version 4.3.3
Libc version: N/A

Python version: 3.12.13 (main, Mar  3 2026, 12:39:30) [Clang 21.0.0 (clang-2100.0.123.102)] (64-bit runtime)
Python platform: macOS-26.5.2-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M2

Dependency Versions:
--------------------
tabpfn: 8.1.0
torch: 2.12.1
numpy: 2.3.5
scipy: 1.16.3
pandas: 2.3.3
scikit-learn: 1.7.2
typing_extensions: 4.15.0
einops: 0.8.2
huggingface-hub: 1.14.0
```

LennartPurucker:
Fixes PriorLabs/tabpfn-extensions#350

## Problem

With the `tabpfn-client` backend, `ClientTabPFNRegressor.predict(..., output_type="full")` imported `FullSupportBarDistribution` from `tabpfn.model.bar_distribution` *before* delegating to the client. That module shim was removed in tabpfn 8.1 — this package's own minimum version — so the call always raised, even though current tabpfn-client already includes a compatible `criterion` in its full output. This broke e.g. `TabPFNUnsupervisedModel.outliers()` on the client backend.

## Fix

* `predict(output_type="full")` now calls the client **first** and passes a client-provided `criterion` through untouched (current tabpfn-client attaches one itself when the local `tabpfn` package is installed).
* Only when an older client returns raw arrays is the criterion reconstructed from the returned `borders`, importing `FullSupportBarDistribution` via `tabpfn.regressor` — the same version-stable path tabpfn-client itself uses (the class' concrete module has moved twice: `tabpfn.model.bar_distribution` → `tabpfn.architectures.base.bar_distribution` → `tabpfn.architectures.shared.bar_distribution`).
* The criterion handling lives in a module-level helper, `_client_full_output_with_criterion`, so it is unit-testable without a tabpfn-client install or API credentials.
* Drive-by fix in the same method: other output types are now actually forwarded to the parent predict, as the docstring already promised. Previously `predict(X, output_type="quantiles", quantiles=grid)` silently dropped both arguments and returned means — `pval_crt` and `cp_missing_data` call it exactly that way.

## Verification (no API credentials needed)

In a fresh venv with `tabpfn 8.1.0` + `tabpfn-client 0.3.3`, running the issue's exact reproduction with the client network layer stubbed:

* **before the fix**: `ValueError: output_type='full' requires the TabPFN package with FullSupportBarDistribution to be installed` (the exact traceback from the issue, despite tabpfn 8.1.0 being installed)
* **after the fix**: `outliers()` returns 16 finite scores, both when the stub mimics the current client (criterion included) and an older client (criterion reconstructed from borders)

New tests in `tests/test_client_full_output.py` pin all of this: criterion pass-through, reconstruction from borders (input dict not mutated, criterion usable), a clear error when the local tabpfn package is missing, and wrapper-level pass-through of `output_type`/`quantiles` (client's network predict monkeypatched). All 5 pass in the fresh venv; the two wrapper tests skip gracefully where tabpfn-client is not installed. `ruff check` and `ruff format --check` pass with the pinned ruff 0.8.6.


oscarkey (commented):
good morning! lgtm, but what do you think about simplifying things by just requiring a new enough `tabpfn-client`, rather than having the workaround?

LennartPurucker:
Would also work for me! 

I am not sure what our policy on backward compatibility is in the extensions repo. Do you have an idea?

oscarkey:
We're aggressive at bumping the tabpfn dependency, not sure if it's different for tabpfn-client.

LennartPurucker:
then let me do the same here

oscarkey (approved):
ah sorry just saw you made the fix. Lgtm!

PR #360, opened 2026-07-24
test: add full example run mode (scheduled GPU); unify example shrinking on FAST_TEST_MODE (PRI-330)

adrian-prior:
Implements [PRI-330](https://linear.app/priorlabs/issue/PRI-330/nightly-gpu-example-tests-run-all-examples-to-completion-strict-make): a scheduled GPU job that runs every example **to completion, at full size, exactly as a user would run it** — where a timeout **is** a failure — plus the harness and example changes that support it.

## Full run mode (`tests/test_examples.py`)

* **Smoke** (default, per-PR): timeout = pass; examples run with `FAST_TEST_MODE=1` so those that support it (directly or via library code) shrink their workload.
* **Full** (`EXAMPLE_FULL=1`): every example runs **to completion** at **full size** — `FAST_TEST_MODE` is stripped from the subprocess env, so the run verifies exactly what a user executes. Timeout (900s; slowest measured example \~300s) = failure. Env-var switch so CLI args can't flip pass/fail semantics.
* One shrink flag: the harness's `TEST_MODE` convention is gone (nothing ever read it); everything uses `FAST_TEST_MODE`, same as the unit suite and the `unsupervised` library paths.
* Example subprocesses default to `MPLBACKEND=Agg` (`plt.show()` on a machine with a display blocks on the GUI window, which the full run would misread as a timeout).

## Shrunk slow examples (smoke-gate coverage)

`feature_selection.py`, `pdp_example.py`, `pval_crt_demo.py` shrink under `FAST_TEST_MODE=1` and complete within the smoke budget (10–43s on CPU) instead of timing out unverified. Full-size behavior unchanged — that's what the full run verifies.

## Scheduled workflow (`test_examples_full.yml`)

Cron every 2 days + `workflow_dispatch`, `ubuntu-22.04-4core-gpu`, sequential, `timeout-minutes: 90` (measured full run: \~17 min). On failure or job timeout it opens or updates a `nightly-failure` tracking issue (pattern from TabPFN's nightly notebook check, plus `cancelled()` so job timeouts alert too).

## Bugs found by the full run itself (fixed here)

1. `bayesian_optimization.py`: botorch objective's bounds buffer stays on cpu while candidates are on cuda → crash on any GPU machine. Fixed by moving the objective to the evaluation device.
2. One level deeper, `bayesian_optimization/bo.py::expected_improvement`: TabPFN builds `raw_space_bardist_` on cpu → EI crashes on cuda. **The extension's whole GPU path was broken**; unnoticed because its unit tests pin `device="cpu"` (follow-up: [PRI-355](https://linear.app/priorlabs/issue/PRI-355/unit-tests-pinning-devicecpu-bypass-gpu-ci-bo-cuda-path-was), stacked as PriorLabs/tabpfn-extensions#364).

## Verified

The workflow was exercised end-to-end on this PR via a temporary trigger (since removed) — see the [run-by-run summary comment](<https://github.com/PriorLabs/tabpfn-extensions/pull/360#issuecomment-5068717097>): final full run **green** (17 passed at the time, before the deprecated phe/hpo examples were removed on main in PriorLabs/tabpfn-extensions#361), issue-filing guard verified, demo issue PriorLabs/tabpfn-extensions#362. After merge: one manual `workflow_dispatch` on `main` as the definitive check.


examples/interpretability/feature_selection.py
```diff
@@ -3,14 +3,23 @@
...
+# the example finishes quickly; results are correspondingly rougher.
+FAST_TEST_MODE = os.environ.get("FAST_TEST_MODE") == "1"
+
 # Load data
 data = load_breast_cancer()
 X, y = data.data, data.target
 feature_names = data.feature_names
+if FAST_TEST_MODE:
+    # Selection cost grows with rows, candidate features, and rounds.
+    X, y, feature_names = X[:100, :10], y[:100], feature_names[:10]
```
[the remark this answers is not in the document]
adrian-prior: Checked: `load_breast_cancer()` is not globally class-ordered — only the first ~19 rows are malignant. `np.bincount(y[:100])` gives `[65, 35]`, so the slice keeps both classes, and the shrunk example runs green through the harness (stratified 5-fold CV, score 0.93). Leaving as is.

adrian-prior:
For reviewers — the new workflow was exercised on this PR via the temporary `pull_request` trigger (removed before merge). The runs document the whole mechanism, and caught two real GPU-only bugs along the way:

1. **[Run 1](https://github.com/PriorLabs/tabpfn-extensions/actions/runs/30083511005)** (first strict run): caught bug #1 — `bayesian_optimization.py` crashed on CUDA with a device mismatch (botorch bounds buffer on cpu vs. inputs on cuda). Invisible to the CPU smoke gate. Timings: full suite **16 min**, slowest example 299s vs. the 900s per-example cap.
2. **[Run 2](https://github.com/PriorLabs/tabpfn-extensions/actions/runs/30084571947)** (failure-notification step added, example still unfixed): failed as expected, and the "Open or update tracking issue on failure" step was **skipped** — the `pull_request` guard works, so PR runs never file issues. Issues are only filed for scheduled/dispatched runs.
3. **[Run 3](https://github.com/PriorLabs/tabpfn-extensions/actions/runs/30085134112)** (example fixed): caught bug #2, one call deeper — `expected_improvement` in `bayesian_optimization/bo.py` mixes devices because TabPFN builds `raw_space_bardist_` on cpu. The extension's whole cuda path was broken: the BO unit tests pin `device="cpu"`, so even the GPU CI job never exercised it.
4. **[Run 4](https://github.com/PriorLabs/tabpfn-extensions/actions/runs/30086400455)** (bo.py device fix): **green** — 17 passed, 3 skipped (deprecated phe/hpo + GPL survival), 1 xfailed (tabebm, #225) in 16m52s.

What an auto-filed failure issue looks like: #362 (manually filed demo, closed).

.github/workflows/test_examples_full.yml
```diff
@@ -0,0 +1,134 @@
...
+                body,
+              });
+            } else {
+              await github.rest.issues.create({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                title,
+                body,
+                labels: ["nightly-failure"],
+              });
```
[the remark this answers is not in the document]
adrian-prior: Moot here: the `nightly-failure` label already exists in this repo (created alongside this PR), and the create-with-label path was exercised end-to-end via the demo issue #362. Same standing assumption as TabPFN's nightly notebook check, which this ports.

.github/workflows/test_examples_full.yml
oscarkey, line 1: Maybe the file name should be `test_examples_full.yml` or `examples_full_test.yml`?

and same for `name:` and `run-name:`

.github/workflows/test_examples_full.yml
```diff
@@ -0,0 +1,131 @@
...
+        env:
+          CUDA_VISIBLE_DEVICES: "0"
+          # Exclude cpu/mps so TabPFN selects the GPU. The cpu code paths are
+          # already covered by the per-PR smoke gate.
+          TABPFN_EXCLUDE_DEVICES: "cpu,cpu:0,mps"
+          EXAMPLE_STRICT: "1"
+        run: |
+          set -o pipefail
+          uv run --no-sync pytest tests/test_examples.py --run-examples -v 2>&1 \
+            | tee pytest_output.txt
```
oscarkey, line 80: maybe also print pytest_output.txt at the end, so we can see what happened in the logs

.github/workflows/example_full_run.yml
```diff
@@ -0,0 +1,131 @@
...
+            | tee pytest_output.txt
+
+      # Scheduled-run failures don't notify anyone by default (GitHub only
+      # emails whoever last edited the cron line), so track them as an issue.
+      # Repeat failures comment on the open issue instead of filing new ones.
+      # cancelled() covers the job-timeout case: exceeding timeout-minutes
+      # cancels the steps rather than failing them, so failure() alone would
+      # stay silent exactly when the run hangs.
+      - name: Open or update tracking issue on failure
+        if: (failure() || cancelled()) && github.event_name != 'pull_request'
```
oscarkey: `&& github.event_name != 'pull_request'` don't think this can ever happen as the trigger is only cron and manual launch. Is it meant to be hardening?

tests/test_examples.py
```diff
@@ -1,22 +1,32 @@
...
+Each example is executed as a subprocess with a fixed per-example timeout.
+There are two run modes:
+
+* **Smoke** (default): an example that *errors* fails the test; one that simply
+  doesn't finish in time *passes* -- we only assert it starts and runs without
+  crashing. Examples run with ``FAST_TEST_MODE=1`` so those that support it
+  (directly, or via library code that reads the flag) shrink their workload.
+  This is the fast, cheap per-PR guard against broken example scripts and
+  import errors.
+* **Strict** (``EXAMPLE_STRICT=1``): every example must run *to completion*,
```
oscarkey: Should we write "FULL" rather than "STRICT" here, to match the terminology we use in the workflow file?

oscarkey (approved):
nice!

adrian-prior:
This change is part of the following stack:

- #360 ◀
    - #364

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

.github/workflows/example_full_run.yml
adrian-prior: Yeah, that was a guard I added for testing. Will remove.

.github/workflows/test_examples_full.yml
adrian-prior: But will this not already be logged in the action's output? My intern tells me this:

> Oscar's "print pytest_output.txt at the end" (#2): I believe it's based on a misreading — tee already streams the full pytest output live into the step log; the file exists only so the failure-issue can quote the FAILED lines. Printing it again would duplicate \~1000 lines in the log.

tests/test_examples.py
adrian-prior: Will rename to full!

PR #363, opened 2026-07-24
Adding a decoder read-out as an interpretability feature

eliott-kalfon:
The `ManyClassDecoder` offers interesting opportunities for interpretability. This PR introduces both a feature to retrieve the raw decoder readout, along with a script to visualise the decoder readouts for a set of test rows.

Note to the reviewer: we could make some changes on TabPFN public to make reading the weights a bit less awkward. Still, I am ok with the current implementation.

image

This leverages the ICL-ness of TabPFN and goes beyond black-box explainability.


examples/interpretability/decoder_readout_example.py
```diff
@@ -0,0 +1,193 @@
...
+# 2D projection of the model's training embeddings (averaged over the ensemble).
+train_emb = clf.get_embeddings(X_test, data_source="train").mean(axis=0)
+query_emb = clf.get_embeddings(X_test[queries], data_source="test").mean(axis=0)
+
+try:
+    from umap import UMAP
+
+    reducer = UMAP(n_components=2, random_state=0)
+    proj_name = "UMAP"
+except ImportError:
```
adrian-prior: nit: Maybe something like a warning here?

examples/interpretability/decoder_readout_example.py
```diff
@@ -0,0 +1,193 @@
...
+def pick_queries(p: np.ndarray) -> list[int]:
+    """Four test rows spanning confident-negative to confident-positive."""
+    order = np.argsort(p)
+    split = np.clip(np.searchsorted(p[order], 0.5), 1, len(order) - 1)
+    return [order[0], order[split - 1], order[split], order[-1]]
+
+
+queries = pick_queries(p_pos)
+
+# 2D projection of the model's training embeddings (averaged over the ensemble).
```
adrian-prior: Quick idea: How about we make the plotting a reusable function, where we feed in the embeddings etc. For example, really like that we reusue the TabPFN-embeddings here, but I would find it really interesting if we could also use default PCA/TSNE embeddings on the original features space (because the feature space oc looks quite seperable when we use the target-conditioned embeddings). This would actually be a great way to show what TabPFN does under the IMO. (i.e. first plot using the features + PCA -> probably the "locality of the multi-class head is way less present", second plot like we have it right now, where the multi-class head basically just uses distance in the embedding space).

src/tabpfn_extensions/interpretability/decoder_readout.py
```diff
@@ -0,0 +1,187 @@
+#  Copyright (c) Prior Labs GmbH 2025.
```
adrian-prior, line 1: Could you have a look at whether this CC comment makes any sense:

> test-row chunking silently corrupts the estimator axis (src/tabpfn_extensions/interpretability/decoder_readout.py:154, the np.concatenate(captured, axis=0)). With fit_mode="fit_with_cache" and more than max_batched_test_rows (default 32768) test rows, tabpfn runs the decoder once per chunk (verified in tabpfn 8.1.0 inference.py:1139), so captured holds per-chunk arrays. Concatenating on axis 0 then either raises a shape error (unequal chunks) or — when chunks divide evenly — silently returns a result where the "estimator" axis is actually test-row chunks and the reported n_test is wrong. A cheap guard would close this: after building weights, assert weights.shape[-2] == len(X) and raise a clear NotImplementedError otherwise. Edge case, but the failure mode is silent wrongness. (The default fit_mode="fit_preprocessors" is unaffected; memory_saving_mode chunks inside layers, not the decoder call.)

adrian-prior (commented):
Overall great PR. I still a few smaller questions/suggestions. 

Also, should the .png file be commited into the repo? If so, where do we actually use it?

examples/interpretability/decoder_readout_example.py
eliott-kalfon: I love the idea. If I understand correctly, we would then have a plotting function that would take as input: optional embeddings (defaults to PCA/UMAP over feature space) and the weights, and plots that

src/tabpfn_extensions/interpretability/decoder_readout.py
eliott-kalfon: This may be true, I am on it. I had a similar issue with subsampling. Will test and report back

eliott-kalfon: Now handled

src/tabpfn_extensions/interpretability/decoder_readout.py
```diff
@@ -0,0 +1,415 @@
...
+    y_train: np.ndarray,
+    class_names: list[str],
+    *,
+    y_test: np.ndarray | None = None,
+    embeddings: tuple[np.ndarray, np.ndarray] | None = None,
+    query_titles: list[str] | None = None,
+    colors: list[str] | None = None,
+    title: str = "TabPFN decoder-head readout",
+    top_k: int = 20,
+) -> Figure:
```
adrian-prior, line 256: Some pedandic CC errors. You can decide whether you want to support them or not (It might be nice to raise with a nice error in the case that we don't think it's worth supporting though):

# PR #363 — two confirmed crashes in `plot_decoder_readout`

**1. `class_names` is indexed by label value while `colors` is indexed by position**
(`decoder_readout.py:375`, `:377`, `:390` vs `:307`). Any labels that aren't
0-indexed integers crash: labels `{1,2}` give `IndexError`, string labels give
`TypeError: list indices must be integers or slices, not numpy.str_`.

**2. `gaussian_kde` dies on any class with fewer than 3 rows**
(`decoder_readout.py:314`). One row raises `ValueError`, two raise `LinAlgError`,
and it's projector-independent — 2 points in 2D always give a rank-1 covariance, so
`umap-learn` doesn't help.

Both verified against `tabpfn 8.1.0`. CI is green because nothing tests
`plot_decoder_readout` and the one example uses well-populated 0-indexed labels.

src/tabpfn_extensions/interpretability/decoder_readout.py
```diff
@@ -0,0 +1,415 @@
...
+their one-hot labels weighted by that attention. The prediction is therefore a
+weighted vote, and ``P(class c)`` for a test row is the sum of its attention
+weights over the training rows whose label is ``c``.
+
+``get_decoder_readout`` recovers those per-training-row attention weights, so you
+can see *which* training points drive a prediction and by how much. For each test
+row the weights sum to 1 (averaged over the decoder's attention heads and over the
+ensemble members). Collapsing them by training label with ``class_vote`` reproduces
+the model's ``predict_proba`` up to the head's log-clamping, at the classifier's
+default ``softmax_temperature=0.9`` and ``balance_probabilities=False``. Both
```
adrian-prior: Maybe, we can quickly sync regarding this offline before we merge. I am really struggling to wrap my head around this by CC is pretty adament about this:

Checked the attention math against tabpfn 8.1.0 and it holds up. One thing on the `predict_proba` claim though.

The identity holds at `softmax_temperature=1.0`, not at the default 0.9. `class_vote` is upstream of the temperature - the head's `log` and the classifier's softmax cancel, but the `÷T` sitting between them has no counterpart, so what comes out is `predict_proba ∝ vote^(1/T)`, a sharpened vote. Measured max gaps:

| classes | T=0.9 (default) | T=1.0 |
|---|---|---|
| 2 | 0.0210 | 0.0000 |
| 5 | 0.0369 | 0.0001 |
| 10 | 0.0566 | 0.0003 |

The T=1.0 column is the log-clamping alone, so at the default the gap is ~190x what the docstring attributes it to.

Two consequences for the test at `:87`:

- `atol=0.05` is 50x looser than log-clamping needs, so it can't catch a real regression in the attention math. Fitting at `softmax_temperature=1.0` lets it drop to `1e-3`.
- Multiclass is untested and would fail: replaying that same assertion, 3 and 5 classes pass but 10 gives 0.0582 > 0.05. `multiclass_data` is sitting unused in `conftest.py:337`.

Suggested docstring change (`:14-19`, `:178-182`, `README.md:24`):

> Collapsing them by training label with `class_vote` reproduces the model's `predict_proba` up to the head's log-clamping when `softmax_temperature=1.0` and `balance_probabilities=False`. Both are applied to the decoder's logits *after* this readout, so at the library default `softmax_temperature=0.9`, `predict_proba` is a sharpened version of the vote (per estimator, ∝ vote^(1/T)), differing by up to ~2pp for binary and ~6pp at 10 classes.

I wouldn't change `class_vote` itself - the raw vote is the right quantity, it's just the claim around it that needs adjusting.

adrian-prior (dismissed):
Looks really good! I left a few small comments. Honstely, I still haven't really wrapped my head around some of the details, so I would be glad if we could chat offline quickly before we merge this (mostly for my own understanding, but could also be helpful for the review)

src/tabpfn_extensions/interpretability/decoder_readout.py
eliott-kalfon: Yes, I have had this issue too. The claim is worth adjusting, but I would go ahead as is. In the next PR (once we release a new TabPFN version) I will make this one a lot more straightforward

src/tabpfn_extensions/interpretability/decoder_readout.py
eliott-kalfon: I mean 1 is valid, I also got this one, on second thought, it makes sense to implement.

src/tabpfn_extensions/interpretability/decoder_readout.py
adrian-prior: Sounds good. it's approved so from my side you can go ahead and merge 🙂

adrian-prior (approved):

PR #364, opened 2026-07-24
test: run device-flexible extension tests on cuda when available

adrian-prior:
Implements [PRI-355](https://linear.app/priorlabs/issue/PRI-355/unit-tests-pinning-devicecpu-bypass-gpu-ci-bo-cuda-path-was): the GPU CI job ran the BO and CRT suites nominally "on GPU" while every TabPFN they constructed was pinned to `device="cpu"` — which is how the BO extension's cuda path could be completely broken (fixed in #360) under a green GPU job.

## Changes

- `tests/conftest.py`: new `TEST_DEVICE` — cuda when available (so the GPU job exercises real cuda paths), cpu otherwise, never mps (CI excludes it; auto-selection would silently switch local macOS runs).
- `tests/test_bayesian_optimization.py`: regressor + all test tensors on `TEST_DEVICE` — this suite's cpu pin is what hid the broken cuda path.
- `tests/test_pval_crt.py`: all four `device="cpu"` → `TEST_DEVICE`.
- `tests/test_interpretability_shapiq.py` deliberately unchanged: its docstring documents the cpu pin as load-bearing (exact fp32 comparisons need deterministic cpu).

## Verification

Local (cpu): all three suites pass unchanged (`TEST_DEVICE` resolves to cpu). The cuda leg is verified by this PR's own `Test on GPU` jobs — with these changes they would have caught the `raw_space_bardist_` bug.

**Stacked on #360** (needs its `bo.py` device fix — the BO tests on cuda fail without it); retargets to `main` once #360 merges.

### DIFF
--- a/tests/test_bayesian_optimization.py
+++ b/tests/test_bayesian_optimization.py
@@ -8,6 +8,7 @@ bo = pytest.importorskip(
     reason="bayesian_optimization requires the full TabPFN package",
 )
 
+from conftest import TEST_DEVICE  # noqa: E402
 from tabpfn import TabPFNRegressor  # noqa: E402
 
 if not hasattr(TabPFNRegressor, "fit_with_differentiable_input"):
@@ -23,7 +24,7 @@ DIM = 3
 @pytest.fixture(scope="module")
 def train_data():
     torch.manual_seed(0)
-    train_x = torch.rand(8, DIM)
+    train_x = torch.rand(8, DIM, device=TEST_DEVICE)
     train_y = train_x.sum(dim=1)
     return train_x, train_y
 
@@ -31,7 +32,7 @@ def train_data():
 def make_regressor() -> TabPFNRegressor:
     return TabPFNRegressor(
         n_estimators=1,
-        device="cpu",
+        device=TEST_DEVICE,
         random_state=0,
         inference_precision=torch.float32,
         differentiable_input=True,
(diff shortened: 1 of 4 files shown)


adrian-prior:
This change is part of the following stack:

- #360
    - #364 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

oscarkey (commented):
in `test_pval_crt.py`, does it work to just leave `device` as "auto"?

Then we only need this in `test_bayesian_optimization.py`

oscarkey (dismissed):

tests/test_bayesian_optimization.py
```diff
@@ -8,6 +8,7 @@
     reason="bayesian_optimization requires the full TabPFN package",
 )
 
+from conftest import TEST_DEVICE  # noqa: E402
 from tabpfn import TabPFNRegressor  # noqa: E402
```
oscarkey: If you move these above `bo` the you can remove the `noqa`

adrian-prior: Done for the conftest import. The tabpfn one can't move above the guard — on client-only setups (no full tabpfn) a top-level import fails collection instead of skipping — so it's now an `importorskip` too, which drops the noqa as well.

adrian-prior:
Re `device="auto"` in `test_pval_crt.py`: yes — `tabpfn_crt`'s own default is already cuda-if-available (never mps), so I dropped the kwarg there entirely. `TEST_DEVICE` stays only in the BO tests: `TabPFNRegressor`'s default `"auto"` would pick mps on a local Mac, and the test tensors need an explicit device anyway.

oscarkey (approved):

oscarkey (approved):

PR #372, opened 2026-08-02
Assign external-PR reviewers deterministically, not randomly

noahho:
Replaces the random reviewer pick with a deterministic one keyed on the PR number. The file stays as the source of reviewers.

```diff
-const reviewer = reviewers[Math.floor(Math.random() * reviewers.length)];
+const reviewer =
+  reviewers[context.payload.pull_request.number % reviewers.length];
```

### Why

Dependabot opens its PRs as a run of consecutive numbers, and `Math.random()` happily gives the same person several in a row - one reviewer drew both #368 and #369 in tabpfn-extensions. Modulo makes consecutive numbers structurally distinct.

Measured against the real eligible-PR history (fork + Dependabot PRs, 8 reviewers):

| | TabPFN (59 PRs) | tabpfn-extensions (101 PRs) |
|---|---|---|
| Back-to-back repeats, modulo | 3 of 58 | 3 of 100 |
| Back-to-back repeats, random | 8 | 13 |
| Consecutive-number pairs landing on the same reviewer | 0 of 27 | 0 of 40 |

**What this is not:** a round robin. Long-run counts are about as uneven as random (modulo spread 4-10 across 8 buckets on TabPFN, 8-16 on tabpfn-extensions; random baseline 4-11 and 8-18), because the PR numbers that reach this workflow are sparse - it fires only for forks and Dependabot, and PR numbers share a counter with issues. Equal counts per reviewer would need stored state or GitHub's own team-based review assignment. This fixes the clustering, not the fairness, and the comment in the workflow says so.

Side benefit: it is idempotent. A reopen or ready-for-review re-request now names the same person instead of possibly adding a second reviewer.

Same change is going into the other copy of this workflow (TabPFN and tabpfn-extensions are the only two repos that have it).

### DIFF
--- a/.github/workflows/assign-pr-reviewer.yml
+++ b/.github/workflows/assign-pr-reviewer.yml
@@ -1,5 +1,5 @@
-# When a PR is opened from a fork or by dependabot, request a review from a randomly
-# chosen reviewer so the PR gets picked up.
+# When a PR is opened from a fork or by dependabot, request a review from a reviewer chosen
+# deterministically from the PR number, so the PR gets picked up.
 #
 # We use this workflow instead of GitHub's built-in auto-assignment system because we
 # want to avoid auto assigning reviewers to PRs from Prior Labs team members, and GitHub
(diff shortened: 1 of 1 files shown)


adrian-prior:
@noahho, happy to review this PR. Seems like you need to rebase onto main, and there are a few comments. Could you request my review once these issues are resolved?

PR #373, opened 2026-08-12
tightened comparison between class vote and predict proba

eliott-kalfon:
Resolving this [thread](https://prior-labs.slack.com/archives/C0AMY58MR62/p1786460838113129)

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -97,9 +97,6 @@ jobs:
           TABPFN_TOKEN: ${{ secrets.TABPFN_TOKEN }}
         run: uv run --no-sync python scripts/download_all_models.py
 
-      # The allow-list and the ignored packages live in [tool.licensecheck] in
-      # pyproject.toml. Passing them here as flags would override that table rather
-      # than add to it.
       - name: Check for forbidden licenses
         if: runner.os == 'MacOS' && matrix.dependency-set == 'lowest-direct'
         shell: bash
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -97,14 +97,15 @@ jobs:
           TABPFN_TOKEN: ${{ secrets.TABPFN_TOKEN }}
         run: uv run --no-sync python scripts/download_all_models.py
 
+      # The allow-list and the ignored packages live in [tool.licensecheck] in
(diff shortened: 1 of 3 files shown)


jmkuebler:
@eliott-kalfon can you briefly describe what the root cause of the previous flakyness were and how its changed now (and why it still test the thing we mean to test). I am struggling to understand that from the diff.

eliott-kalfon:
> @eliott-kalfon can you briefly describe what the root cause of the previous flakyness were and how its changed now (and why it still test the thing we mean to test). I am struggling to understand that from the diff.

Sure! In TabPFN, we convert vote to logits using the [following](https://github.com/PriorLabs/TabPFN/blob/main/src/tabpfn/architectures/tabpfn_v3.py#L571):

`torch.log(torch.clamp(test_output_MBT, min=1e-5) + 3e-5)`

This is done at the estimator level. `predict_proba` then applies softmax over these.

In the [previous version of the test](https://github.com/PriorLabs/tabpfn-extensions/pull/373/changes#:~:text=weights%2C%20_%20%3D%20get_decoder_readout(clf%2C%20X_test)), I retrieved the weights at the ensemble level (for simplicity).

This is an issue as the "log clamping + adding small constant" does not happen at the same level, which introduced an error, especially in the multiclass setting, with many probabilities close to 0. This could bring the delta on either side of the assigned tolerance (1e-3).

I now generated a bit of code to retrieve the probabilities at the estimator level, to match what happens in the `predict_proba` - after this, I could even make the test tighter to 1e-6.

I am still comparing the outcome of `predict_proba` and the attention recomputation of the `ManyClassDecoder`. Note that this problem will vanish after the next TabPFN release, in which we will just be able to get this readout without recomputation.

jmkuebler (dismissed):
Thank you for the additional context! LGTM

eliott-kalfon:
Rerequesting review as I fixed the CI issue due to the fsspec license check (see [thread](https://prior-labs.slack.com/archives/C0AMY58MR62/p1786552658561139)) and [PR](https://github.com/PriorLabs/TabPFN/pull/1160/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711) in TabPFN

jmkuebler (approved):
LGTM, thank you!

PR #382, opened 2026-08-19
Fix bf16-dependent flake in the decoder-readout class-vote test

bejaeger:
`test_interpretability_decoder_readout.py::test_class_vote_matches_predict_proba` fails on some CI runs and not others, with the same code and the same resolved dependencies — most recently on `ubuntu-latest, 3.13, highest` in #380, where a re-run of the identical job passed.

## Cause

The test asserts an identity: collapsing the readout weights by training label reproduces `predict_proba`, because the head builds its logits as `log(clamp(vote, min=1e-5) + 3e-5)` and the softmax at `softmax_temperature=1.0` undoes the log. That round trip is only exact in full precision.

With `inference_precision="auto"` (the default) tabpfn enables **bf16 autocast on CPU** whenever the CPU has native bf16 support — Intel AMX / AVX512-BF16 or AMD Zen 4+, see `tabpfn.utils.infer_autocast_inference_mode` → `_cpu_supports_fast_bf16`. In bf16 the log/softmax round trip loses ~1e-2 relative accuracy, far past the check's `atol=1e-6`. GitHub's `ubuntu-latest` fleet is mixed, so whether the head ran in bf16 or float32 — and therefore whether the tight check passed — came down to which CPU the runner happened to land on. macOS runners are unaffected (arm64, no oneDNN fast bf16 kernels, so autocast stays off).

Failure signature from the #380 run: max |Δ| 3.0e-4 binary / 3.1e-3 multiclass, max relative 8.3e-3 / 1.5e-2.

## Fix

Pin the fixture to `inference_precision=torch.float32` rather than loosening the tolerance to bf16 slack. Widening `atol` to ~2e-2 would keep the test green but stop it from catching what it exists to catch: a row-misalignment or an upstream drift between `ManyClassDecoder.attention_weights` and the fused `forward`. Pinning the precision keeps the guard exact and makes it deterministic across runners.

Also added:
- a fixture assertion that autocast is in fact off, so the check fails loudly rather than silently going back to flaking if a forced dtype ever stops disabling autocast upstream;
- a note in the user-facing `class_vote` docstring that the identity is exact only in full precision, since this affects anyone comparing a readout against `predict_proba` on bf16 hardware or on GPU (fp16).

## Testing

Reproduced the failure locally by simulating a bf16-capable CPU (patching `tabpfn.utils._cpu_supports_fast_bf16` to return `True`) — both parametrizations fail exactly as CI did, with max |Δ| 9.5e-4 / 1.4e-3. With this change, all 7 tests in the file pass under the same simulation, and the identity gap in float32 is 1.2e-7. Those numbers were bit-identical on tabpfn 8.3.0 and 8.4.0, so the flake predates the version bump in #380.

Full suite locally: unchanged from before this PR.


eliott-kalfon (approved):
LGTM, thanks for taking this Ben :)

PR #383, opened 2026-08-20
Reapprove internal PRs after a push

klemens-floege:
Generated by Claude Code

Ports `.github/workflows/reapprove-internal-prs.yml` from [PriorLabs/TabPFN](https://github.com/PriorLabs/TabPFN/blob/main/.github/workflows/reapprove-internal-prs.yml), byte-identical — this repo already uses `ubuntu-slim` for its bot workflows, so nothing needed adjusting.

Branch protection dismisses every approval when a PR is pushed to. That check earns its keep for external contributors, but on internal PRs it means an approval evaporates on each rebase or fixup. The workflow restores an approval that was dismissed *by a push*, and only when the PR author, the dismissed reviewer, and the branch are all internal.

<details><summary>Guards before it approves anything</summary>

- PR author is `MEMBER` or `OWNER`
- Dismissed review came from a `MEMBER`/`OWNER` (or the bot's own earlier approval), so an outside approval can't be used to trigger it
- Head branch is in this repo, not a fork
- The dismissed review was actually an *approval* — not a changes-requested
- `dismissal_commit_id` is set, so a reviewer who dismisses by hand is never overridden
- Exits early if the current head is already approved

Upstream the job has run 7 times and been gated out (`skipped`) every time, so the YAML — including the `concurrency: queue: max` key — is known to parse and run.
</details>

CI-only, no user-facing change, so it takes the `no changelog needed` label per `changelog/README.md`.

### DIFF
--- a/.github/workflows/reapprove-internal-prs.yml
+++ b/.github/workflows/reapprove-internal-prs.yml
@@ -1,0 +1,116 @@
+# Automatically reapprove PRs from PriorLabs org members after new code is pushed.
+#
+# The branch protection rules dismiss all approvals when new code is pushed to a PR, to
+# ensure the most recent push is reviewed and approved. This is important for external
+# collaborators, but is extra overhead for internal PRs where we trust the authors. In
+# this case, the workflow automatically approves the current version of the PR if a
+# previous version was already approved.
+name: Reapprove internal PRs
+
+on:
+  pull_request_review:
+    types: [dismissed]
+
+permissions:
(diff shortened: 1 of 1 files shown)


oscarkey (commented):
Let's hold off for a bit because this workflow doesn't actually work yet. It's on my list to fix.

oscarkey:
I copied the fixed version in https://github.com/PriorLabs/tabpfn-extensions/pull/386

PR #386, opened 2026-08-21
Copy automatic PR reapproval workflow from TabPFN.

oscarkey:
See
https://github.com/PriorLabs/TabPFN/blob/main/.github/workflows/reapprove-internal-prs.yml

`secrets.PRIORLABS_ORG_MEMBERSHIP_READ_PAT` is set at the org level, and made available to this repo.

### DIFF
--- a/.github/workflows/reapprove-internal-prs.yml
+++ b/.github/workflows/reapprove-internal-prs.yml
@@ -1,0 +1,141 @@
+# Automatically reapprove PRs from PriorLabs org members after new code is pushed.
+#
+# The branch protection rules dismiss all approvals when new code is pushed to a PR, to
+# ensure the most recent push is reviewed and approved. This is important for external
+# collaborators, but is extra overhead for internal PRs where we trust the authors. In
+# this case, the workflow automatically approves the current version of the PR if a
+# previous version was already approved.
+name: Reapprove internal PRs
+
+on:
+  pull_request_review:
+    types: [dismissed]
+
+permissions:
+  pull-requests: write
+
+# One push can dismiss several approvals, and each dismissal starts its own run. Run
+# them one at a time per PR, so that a later run sees the approval an earlier one
+# created instead of adding a duplicate.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: false
+  # We need to check every review that was dismissed, in case it was an approval.
+  queue: max
+
+jobs:
+  reapprove:
+    name: Restore dismissed approval
+    # The PR does not come from a fork. This also means the author has write access to
+    # the repository, as otherwise they could not have pushed the branch. The next step
+    # checks the author's org membership as well, for defence in depth.
+    if: github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-slim
+    steps:
+      - name: Check the author and the reviewer are org members
+        id: membership
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          # We cannot read the memberships with GITHUB_TOKEN.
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -137,12 +140,15 @@ jobs:
           $env:FAST_TEST_MODE = 1
           uv run --no-sync pytest tests/
 
-      - name: Save model cache
-        if: github.ref == 'refs/heads/main'
+      - &save-data-cache
```
klemens-floege: Codex is concerned here

```
[P1] Every PR can create over 10 GB of isolated caches. The save step now runs in every matrix job and on PRs (lines 143–152
    (https://github.com/PriorLabs/tabpfn-extensions/blob/54b443e23c3250022e61c629b8cbc1f79d76b604/.github/workflows/pull_request.yml#L143-L152)). This PR already created four
    refs/pull/386/merge caches of ~2.81 GB each—11.23 GB total. GitHub says PR caches are usable only by reruns of that PR, not main or sibling PRs, and warns that such
    caches can cause eviction/thrashing; the default repository limit is 10 GB. GitHub cache documentation
    (https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching). Split the caches: keep the large model-cache save restricted to main, and
    independently save the much smaller sklearn cache on PR failures if desired.
```

.github/workflows/reapprove-internal-prs.yml
```diff
@@ -0,0 +1,141 @@
...
+                return true;
+              } catch (error) {
+                if (error.status === 404) {
+                  return false;
+                }
+                throw error;
+              }
+            }
+
+            const author = context.payload.pull_request.user.login;
```
klemens-floege, line 56:
```
- [P1] The bot can approve a commit from an unverified pusher. Membership checks cover only the PR author and dismissed reviewer (lines 56–72
    (https://github.com/PriorLabs/tabpfn-extensions/blob/54b443e23c3250022e61c629b8cbc1f79d76b604/.github/workflows/reapprove-internal-prs.yml#L56-L72)), while createReview
    omits commit_id and therefore approves the live HEAD (lines 128–141
    (https://github.com/PriorLabs/tabpfn-extensions/blob/54b443e23c3250022e61c629b8cbc1f79d76b604/.github/workflows/reapprove-internal-prs.yml#L128-L141)). A non-org
    collaborator with repository write access could push to an internal author’s branch and receive automatic approval; a concurrent push can also land between validation and
    approval. The workflow needs to bind approval to both a verified pusher and a verified SHA. GitHub confirms an omitted commit_id defaults to the PR’s latest commit.
    Review API documentation (https://docs.github.com/en/rest/pulls/reviews#create-a-review-for-a-pull-request).
```

klemens-floege (commented):
I am sligthly out of my depth here, codex is a bit concerned about the cache + bot approvals, would you mind taking a look? Happy to approve if you think its fine

.github/workflows/pull_request.yml
oscarkey: sorry Klemens, this is from a different PR, I messed up my stacking. Now reverted.

.github/workflows/reapprove-internal-prs.yml
oscarkey: This is true, but I think it's actually what we want:
1. Person A creates the PR
2. Person B approves
3. Person A pushes some changes
4. Bot starts, triggered by (3)
5. Person C quickly also pushes some changes
6. Bot approves, including (5) in the approval

Note that all three of these people need to have write access to the repository.

Technically the comment in the workflow is wrong: it says it reapproves for priorlabs org members, whereas this also reapproves for anyone with write access. But, in practice, everyone with write access is an org member.

klemens-floege (approved):
Cool!

PR #390, opened 2026-08-27
fix(unsupervised): accept array logits in outlier detection

ggprior:
TL;DR: matches other call sites. this is for tabpfn-client compatibility which can only return np.array, we need to support both here.
https://prior-labs.slack.com/archives/C09F8DP5K2S/p1787843262502949

<details>
<summary>Claude generated description</summary>

`TabPFNUnsupervisedModel.outliers()` fails on the client backend:

```
AttributeError: 'numpy.ndarray' object has no attribute 'clone'
```

`outliers_single_permutation_` calls `pred["logits"].clone()`, which holds only
for the local package — the local regressor types `logits` as a `torch.Tensor`,
the client returns a numpy array. The sibling call site in
`sample_from_model_prediction_` already handles both, so this mirrors it and
takes `dtype`/`device` from the converted tensor rather than the raw value.

Found by running the TabPFN demo notebook end to end against both backends; the
outlier-detection cell is the first place the difference bites.

The new test stubs the regressor, so it needs no credentials and no network.
Reverting the one-line fix turns it red with the `AttributeError` above.

</details>

### DIFF
--- a/changelog/390.fixed.md
+++ b/changelog/390.fixed.md
@@ -1,0 +1,1 @@
+Outlier detection no longer fails with `'numpy.ndarray' object has no attribute 'clone'` when the unsupervised model runs on the TabPFN client.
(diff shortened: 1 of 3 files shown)


eliott-kalfon (approved):
LGTM :)

PR #395, opened 2026-09-11
Add automatic hurdle regression

adrian-prior:
Generated by Codex

Add `AutoHurdleRegressor` for regression targets with a point mass at zero. It fits a zero/nonzero classifier and a regressor on positive rows, then combines their predictions into means, medians, or quantiles. A runnable Facebook Comment Volume example compares means with RMSE first, then medians with MAE, against plain TabPFN and constant baselines.

Automatic selection uses a configurable training zero-rate threshold. Supplied estimators and the interpolation grid are configurable; the default prediction is the mixture median, using a 21-point grid from 0.01 to 0.99. All-zero targets predict zero without fitting either stage, and signed targets use ordinary regression in automatic mode.

Validation: all 32 hurdle tests passed with local TabPFN 8.4.0, including a real-model integration test. Repository Ruff lint/format checks and pinned pre-commit hooks passed.

<details>
<summary>Example command and captured output</summary>

The example samples 1,000 training rows and 1,000 test rows separately from the published splits, using seed 0. Results depend on the sample and model version.

```bash
python examples/hurdle/facebook_comments.py --n-train 1000 --n-test 1000 --seed 0
```

```text
Rows: 1000 train, 1000 test
Zero targets: 52.7% train, 34.9% test
Hurdle enabled: True
Constant mean        RMSE: 121.6932
TabPFN mean          RMSE: 102.2148
Hurdle mean          RMSE: 98.3218
Constant median      MAE: 32.1650
TabPFN median        MAE: 25.3989
Hurdle median        MAE: 24.8174
```

Data: [Singh, K. (2015), Facebook Comment Volume](https://doi.org/10.24432/C5Q886), CC BY 4.0.

</details>


src/tabpfn_extensions/hurdle/hurdle.py
```diff
@@ -0,0 +1,196 @@
...
+    for an all-zero target with the hurdle enabled.
+    """
+
+    def __init__(
+        self,
+        classifier: Any = None,
+        regressor: Any = None,
+        *,
+        hurdle: Literal["auto"] | bool = "auto",
+        zero_threshold: float = 0.5,
```
eliott-kalfon: I may have missed it in the description, but did we do any tests on the thresholds? At 0.5, is the hurdle model better? Or is it more at 0.8?

eliott-kalfon (dismissed):
lgtm just a quick question on how 0.5 was selected. Thanks!

src/tabpfn_extensions/hurdle/hurdle.py
adrian-prior: Not really. I think this also heavily depends on the dataset (if hurdle modelling is good or not). In general, I think any dataset which a significant distribution of zero targets should benefit from hurdle modelling -> I changed the threshold to 25% for now.

eliott-kalfon (approved):

PR #398, opened 2026-09-14
Add the image extension's example and documentation

alanprior:
Layer 4 of 4 in stack #402, on #400: what surrounds the image extension. #399 adds the `image` extra, #401 the encoder internals, #400 the transformer and estimator wrapper.

## What is in it

- `examples/image/tabpfn_with_images.py` fits the Amazon Bestseller task of MulTaBench, the log price of a best-selling product from its photo, rank, page and ratings, three ways: numeric fields only, picture only, both. It reports RMSE. Under `FAST_TEST_MODE=1` it runs on a 300-row sample.
- `examples/image/README.md` explains the cell format, the gated encoder license and how `TabPFNWithImages` is used.
- `ImageTransformer` and `TabPFNWithImages` are exported from the package root, and one line each goes into `README.md` and `examples/README.md`.
- `tests/test_examples.py` registers the example under `REQUIRES_MODULE`, so it skips unless kagglehub is installed by hand. The archive comes from Kaggle and the encoder's weights are gated on the Hugging Face Hub, so it cannot run in CI without an account on each side.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #400, opened 2026-09-16
Add ImageTransformer and TabPFNWithImages: declared image columns become DINOv3 features

alanprior:
## TL;DR

`ImageTransformer` is a scikit-learn transformer that replaces each declared image column of a DataFrame, cells holding file paths, base64, bytes or PIL images, with 30 float features by default: the frozen DINOv3 CLS embedding, standardised and PCA-reduced on the training rows, appended as `<label>_img_<k>` and suffixed when a name is taken. `TabPFNWithImages` wraps a TabPFN estimator, local or client, runs that transformer at fit and predict and shifts the estimator's `categorical_features_indices` to the expanded frame, so TabPFN only ever sees numbers. Columns are declared by position; a missing or undecodable cell is refused naming the row, and the fitted model pickles without the encoder.

Layer 3 of 4 in stack #402, on #401: the transformer and the estimator wrapper. #398 adds the example and docs.

## What & why

Declared image columns of a DataFrame become TabPFN features, so a table with a picture per row is fitted without leaving the scikit-learn API. Nothing is detected: only the columns named in `image_features_indices` are read as images.

## How it works

- `ImageTransformer`, a scikit-learn transformer. Each cell of a declared column holds one image, as a base64 string, as the image file's path or bytes, or as a PIL image. `fit` replaces the column by `n_components` numeric features, 30 by default: the CLS embedding of a frozen DINOv3 ViT-S/16 through a `Pipeline` of `StandardScaler` and `PCA` fit on the training rows, one pipeline per image column in `reducers_`. `transform` applies the same encoder and pipelines. The output frame keeps the other columns, appends the features as `label_img_i`, carries the labels `get_feature_names_out` reports and the caller's row index. A feature label a kept column already bears, or that two declared columns sharing a label would produce twice, gets `_0`, `_1`, ... appended until it is free. A missing or undecodable cell is refused, naming the column and the row.
- `TabPFNWithImages` wraps an estimator. It moves the estimator's `categorical_features_indices` to the expanded frame, runs the encoder on the estimator's device, fits a clone on the expanded frame and delegates `predict`, `predict_proba` and `score`; `is_classifier` follows the wrapped estimator. The estimator only ever sees numbers, so the local package and the client both work.
- The decoding and the encoder come from #401. The encoder is never stored on a fitted transformer, so fitted models pickle without the weights.

## How to use

```python
from tabpfn_extensions import TabPFNClassifier
from tabpfn_extensions.image import TabPFNWithImages

model = TabPFNWithImages(
    TabPFNClassifier(categorical_features_indices=[1]),
    image_features_indices=[3],  # positions in X, like categorical_features_indices
)
model.fit(X_train, y_train)
proba = model.predict_proba(X_test)
```

The DINOv3 weights are gated on the Hugging Face Hub: accept the license once on the model page and log in with `hf auth login` or `HF_TOKEN`.

## Design decisions

- One encoder only, DINOv3 ViT-S/16, with no `model_name` parameter.
- Images go through the model's own image processor from transformers, as in MulTaBench; PIL decodes them and caps the longest side at 512 pixels first.
- Columns are declared by position, like `categorical_features_indices`; a declaration that names no column is refused, and every image column becomes exactly `n_components` features.

Supersedes PriorLabs/TabPFN#1260.


src/tabpfn_extensions/image/image_transformer.py
```diff
@@ -0,0 +1,172 @@
...
+            blocks[position] = reducers[position].transform(embeddings)
+        # Assigned together at the end, so a failed fit leaves no half-fitted state.
+        self.reducers_ = reducers
+        self.feature_names_in_ = np.asarray([str(c) for c in X.columns], dtype=object)
+        return self._assemble(X, blocks)
+
+    def transform(self, X: pd.DataFrame) -> pd.DataFrame:
+        """Replace the declared columns of `X` by the features fit at training."""
+        check_is_fitted(self)
+        if X.shape[1] != len(self.feature_names_in_):
```
RichaShiny, line 102: Could we validate the fitted DataFrame column order here, not only the number of columns?

There seems to be a silent feature-swap case. For example, fit on [age, rating, photo] and then transform [rating, age, photo]. The image column is still at the expected position, so embedding succeeds, but _assemble() keeps the reordered values and subsequently relabels them using the fit-time feature names. The wrapped estimator can therefore receive rating values under age and vice versa without an error.

Since feature_names_in_ is already stored, checking the exact column sequence, or explicitly reordering to the fitted schema, would prevent this. A regression test with reordered non-image columns would also catch it.

alanprior: Thanks @RichaShiny ! Your comment is technically correct, but I would expect the OS package to handle that. **I need to check this is true, as I might be ruining something accidently.** But unless I'm mistaken, this is something that can happen in our internal `fit` for `tabpfn`, even without images. While I agree that it might not be the best practice in the world to rely on the downstream to behave in a specific way, here I feel it's reasonable. WDYT?

(I'll test this, also on the OS package)

alanprior: Ok, seems you were right, great catch. Fixing this + adding tests.

src/tabpfn_extensions/image/image_transformer.py
```diff
@@ -0,0 +1,179 @@
...
+
+DEFAULT_N_COMPONENTS = 30
+
+
+@set_extension("image")
+class ImageTransformer(TransformerMixin, BaseEstimator):
+    """Replaces each declared image column of a DataFrame by numeric features.
+
+    Args:
+        image_features_indices: Positions in `X` whose cells hold images.
```
adrian-prior: nit: maybe add a sentence regarding how these columns actually look like (what dtype) and what format (i.e. is this only base64 or also file-paths etc.). Alternatively refer to a place where people can read up what it means to be an "image column"

src/tabpfn_extensions/image/image_transformer.py
```diff
@@ -0,0 +1,179 @@
...
+        self, X: pd.DataFrame, y: Any = None, **fit_params: Any
+    ) -> pd.DataFrame:
+        """Fit every declared column and return the expanded frame."""
+        del y, fit_params
+        positions = self._declared_positions()
+        if positions[-1] >= X.shape[1]:
+            raise ValueError(
+                f"`image_features_indices` names {positions}, but X has "
+                f"{X.shape[1]} columns."
+            )
```
adrian-prior: I feel like it's a bit weird that the <0 check is in delared_positions but the >= X.shape[1] check is out here.

src/tabpfn_extensions/image/image_transformer.py
adrian-prior: So we have `feature_names_in` now, but I can see where we would actually check that this matches?

tests/test_image.py
```diff
@@ -0,0 +1,580 @@
...
+    ) -> None:
+        """The premise of the next test: TabPFN itself checks the feature names."""
+        X, y = _estimator_data(regression=False)
+        X = pd.DataFrame({"num": X["num"], "more": X["num"] * 2})
+
+        tabpfn_classifier.fit(X, y)
+
+        with pytest.raises(ValueError, match="feature names should match"):
+            tabpfn_classifier.predict(X[["more", "num"]])
+
```
adrian-prior, line 559: nit: Why are these tests excluding the client?

src/tabpfn_extensions/image/image_transformer.py
```diff
@@ -0,0 +1,179 @@
...
+            except ValueError as e:
+                raise ValueError(f"Column {position}, row {row}: {e}") from e
+            if source is None:
+                raise ValueError(f"Column {position}, row {row}: no image.")
+            sources.append(source)
+        return encode_images(sources, device=self.device, batch_size=self.batch_size)
+
+    def _image_feature_names(self) -> list[str]:
+        return [
+            f"{self.feature_names_in_[i]}_img_{k}"
```
adrian-prior: can we just note here that this isn't perfectly save? I.e. if I have two columns that have the same label (which seeminlgy can occur) or if I have a column "photo_img_0" etc. I think it's overkill to make this nicer but noting the edge case could be helpful

adrian-prior (dismissed):
LGTM! Left some minor comments!

src/tabpfn_extensions/image/image_transformer.py
alanprior: Added

src/tabpfn_extensions/image/image_transformer.py
alanprior: Yeah I agree, probably a weird leftover. Done

src/tabpfn_extensions/image/image_transformer.py
alanprior: You are right and also saw that, but CC said that since we use indices this is actually legal. Anyway I agree it's odd and might have undesirable results, so fixed.

PR #401, opened 2026-09-16
Add the DINOv3 image encoder and the decoding of image cells

alanprior:
## TL;DR

Two private modules behind the image extension: `_preprocessing.py` turns an image cell, given as base64, bytes, a file path or a PIL image, into a decoded RGB image, and `_embeddings.py` embeds images with a frozen DINOv3 ViT-S/16, one float32 CLS vector per image. The encoder loads once per process behind `functools.cache`, decodes and embeds one batch at a time so a long column never holds every image, and picks its device through the new `utils.infer_torch_device`, which also works when TabPFN itself runs through the client. The only public names are `GatedEncoderError` and `image_to_bytes`; #400 wires these modules into the sklearn transformer and the TabPFN wrapper.

Layer 2 of 4 in stack #402, on #399: the two modules the image extension is built on, reviewable on their own. #400 adds the transformer and estimator wrapper on top, #398 the example and docs.

## How it works

- `_preprocessing.py`: `cell_to_image_source` reads a cell, a base64 string (a `data:image/...;base64,` prefix is fine), the path of the image file as a string or a `Path`, the image file's bytes, or a PIL image taken as is; `open_images` decodes bytes, reads paths, copies PIL images, converts to RGB and caps the longest side at 512 pixels, the cap MulTaBench uses; `image_to_bytes` encodes a PIL image as a file, for users who want bytes in their table.
- `_embeddings.py`: `get_dino_encoder(device)` loads `facebook/dinov3-vits16-pretrain-lvd1689m` and its image processor once per process, behind a `functools.cache`, and returns them as a `DinoEncoder`; `encode_images` decodes and embeds the images one batch of `batch_size` at a time, so a long column never holds every decoded image, and returns the CLS token of each image as float32, as MulTaBench does. A gated Hub repo surfaces as `GatedEncoderError` with the license instructions.
- The only encoder is DINOv3; there is no model parameter.

## Design decisions

- A string cell that names an existing file is a path, absolute or relative to the working directory; any other string is base64. A path is read only when its batch is decoded, so a column of paths costs no more memory than the paths. There is no root-folder parameter: a table of file names next to a folder becomes a column of paths in one line.
- PIL decodes, then the model's own image processor resizes and normalises: MulTaBench's recipe, with no preprocessing of our own. The processor is whatever transformers picks for the checkpoint, the torchvision-backed one. An EXIF orientation tag is not applied either, as in MulTaBench; an image is embedded as stored.
- The encoder runs in this process even when TabPFN itself runs through the client, so its device comes from the new `utils.infer_torch_device`, not from `infer_device`, which stands in a CPU for the client: TabPFN's own reading of `device` when the local package is installed, and the same rule on torch directly when it is not, CUDA, else MPS, else the CPU, minus what `TABPFN_EXCLUDE_DEVICES` names; of several devices it takes the first.


tests/test_image_embeddings.py
```diff
@@ -0,0 +1,118 @@
...
+            types.SimpleNamespace(AutoModel=FakeAuto, AutoImageProcessor=FakeAuto),
+        )
+        monkeypatch.setattr(_embeddings, "_ENCODER", None)
+
+        first_model, first_processor = _embeddings.get_dino_encoder(torch.device("cpu"))
+        model, processor = _embeddings.get_dino_encoder(torch.device("cpu"))
+
+        assert model is first_model
+        assert processor is first_processor
+        assert loaded == [IMAGE_ENCODER_MODEL, IMAGE_ENCODER_MODEL]
```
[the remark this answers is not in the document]
alanprior: Solved

src/tabpfn_extensions/image/_preprocessing.py
```diff
@@ -0,0 +1,98 @@
...
+        except (
+            OSError,
+            ValueError,
+            SyntaxError,
+            PIL.Image.DecompressionBombError,
+        ) as e:
+            raise ValueError(f"row {row}: {e}") from e
+        if image.mode == "P":
+            image = image.convert("RGBA")
+        image = image.convert("RGB")
```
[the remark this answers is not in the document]
alanprior: Sounds like an edge-case we don't need to support

src/tabpfn_extensions/image/_preprocessing.py
```diff
@@ -0,0 +1,113 @@
...
+            text = text.split(",", 1)[-1]
+        try:
+            return base64.b64decode("".join(text.split()), validate=True)
+        except (binascii.Error, ValueError) as e:
+            raise ValueError(f"not base64: {e}") from e
+    if cell is None or (pd.api.types.is_scalar(cell) and pd.isna(cell)):
+        return None
+    if _is_pil_image(cell):
+        return cell
+    raise ValueError(f"unsupported cell type {type(cell).__name__}")
```
adrian-prior: I think we should definitely also support reading directly from paths. Coming from CV it's quite uncommon to hand around all this large data in a materialized form. You would try to keep the data format at paths as long as possible to reduce the memory footprint of your scripts.

src/tabpfn_extensions/image/_embeddings.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class DinoEncoder(NamedTuple):
+    """The frozen encoder and the image processor that prepares its input."""
+
+    model: DINOv3ViTModel
+    processor: DINOv3ViTImageProcessor
+
+
+_ENCODER: DinoEncoder | None = None
```
adrian-prior: Can we switch to something like this? Feels cleaner to me. WDYT?

```
@functools.cache
def _load_encoder() -> DinoEncoder:
    _raise_if_no_encoder_dependencies()
    ...
    return DinoEncoder(model, processor)

def get_dino_encoder(device: torch.device) -> DinoEncoder:
    encoder = _load_encoder()
    return DinoEncoder(encoder.model.to(device), encoder.processor)
```

src/tabpfn_extensions/image/_embeddings.py
```diff
@@ -0,0 +1,124 @@
...
+    Raises:
+        ValueError: When `batch_size` is not a power of two, or naming the row
+            whose bytes are not an image.
+    """
+    if batch_size < 1 or not math.log2(batch_size).is_integer():
+        raise ValueError(f"`batch_size` must be a power of two, got {batch_size}.")
+    images = open_images(sources)
+    torch_device = _torch_device(device)
+    model, processor = get_dino_encoder(torch_device)
+    chunks = []
```
adrian-prior: Why not move this also into the batching? Could significantly reduce the memory footprint.

src/tabpfn_extensions/image/_embeddings.py
```diff
@@ -0,0 +1,124 @@
...
+    def __init__(self) -> None:
+        super().__init__(
+            f"The weights of `{IMAGE_ENCODER_MODEL}` are gated on the Hugging Face "
+            f"Hub. Accept the license once at https://huggingface.co/"
+            f"{IMAGE_ENCODER_MODEL}, then log in with `hf auth login` or set "
+            "`HF_TOKEN`."
+        )
+
+
+def _torch_device(device: Any) -> torch.device:
```
adrian-prior: Seems like a valid point by Claude:

Summary: infer_device returns CPU whenever the local tabpfn package is not installed, because TabPFN then runs remotely through the client. The DINOv3 encoder always runs locally, so with the client backend it lands on the CPU even when a GPU is available, with no warning. Ask whether that is intended, or whether the encoder should resolve its device from torch directly.

adrian-prior (commented):
Very cool effort! I left a few comments and I think we can go from there :)

src/tabpfn_extensions/image/_preprocessing.py
alanprior: Yeah, I was a bit hesitant about this, but I think you are right. I'll add support into this. I guess that this can have a few flavors though. One common approach would be to have the filename (e.g. cat.png, dog.png) on the CSV but have the folder itself as a path (/users/alan/my_best_pictures/pets_i_like). I think though that this will make things too complicated, adding one more parameter, also making this col dependent.... Are you OK with us supporting only "global paths"?

src/tabpfn_extensions/image/_embeddings.py
alanprior: I considered this, but this might entangle CPU vs GPU in a way. We could use a dataloader I guess to manage it properly, but I felt this could be a bit of an overkill. Still, might be easy to support actually and perhaps cleaner. Let me see if the DataLoader approach makes sense.

src/tabpfn_extensions/image/_embeddings.py
alanprior: Oh, I completely missed this option to be honest. I forgot that we have those two modes. Thanks!

src/tabpfn_extensions/image/_preprocessing.py
alanprior: Handled this [here](
https://github.com/PriorLabs/tabpfn-extensions/pull/401/changes/b8ab1844e12f7913068a7cf2dc18ccedeea45b96). The user is allowed to pass a path, as long as it is the full/relative path. 

If all pics are stored in the folder "/users/alan/my_best_pictures/pets_i_like", the paths in the csv should be "/users/alan/my_best_pictures/pets_i_like/cat.jpg", not "cat.jpg".

Thanks!

src/tabpfn_extensions/image/_embeddings.py
alanprior: Agreed, much cleaner. Solved [here](https://github.com/PriorLabs/tabpfn-extensions/pull/401/changes/a285df8c9f5b50120aff1a866be579a9ff656bf0).

src/tabpfn_extensions/image/_embeddings.py
```diff
@@ -0,0 +1,123 @@
...
+        device: Where the encoder runs, as TabPFN's `device` argument; `None`
+            means `"auto"`. The encoder runs in this process even when TabPFN
+            itself runs through the client.
+        batch_size: Images per forward pass, a power of two.
+
+    Raises:
+        ValueError: When `batch_size` is not a power of two, or naming the row
+            whose bytes are not an image.
+    """
+    if batch_size < 1 or not math.log2(batch_size).is_integer():
```
adrian-prior: nit: Why do we need this? Maybe a nudge would be enough? I think batching with non-power of two batch sizes also should work, shouldn't it?

src/tabpfn_extensions/image/_embeddings.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def get_dino_encoder(device: torch.device) -> DinoEncoder:
+    """The encoder moved to `device`, and its image processor."""
+    encoder = _load_encoder()
+    return DinoEncoder(encoder.model.to(device), encoder.processor)
+
+
+def encode_images(
+    sources: Sequence[bytes | Image],
```
adrian-prior: nit: Type hint should be update to represent paths as accepted types too.

adrian-prior (dismissed):
LGTM! Left some minor comments.

